PR #9, opened 2022-03-21
fix from/to currency detection for coinbase pro swaps

houqp:


src/dali/plugin/input/rest/coinbase_pro.py
```diff
@@ -311,6 +311,7 @@ def _process_fills(self, transaction: Any, in_transaction_list: List[InTransacti
                 to_currency_size = from_currency_size * RP2Decimal(fill[_PRICE])
                 to_currency_price = usd_volume / to_currency_size
             elif fill_side == _BUY:
+                (from_currency, to_currency) = (to_currency, from_currency)
```
houqp, line 314: this is the one line fix

src/dali/plugin/input/rest/coinbase_pro.py
```diff
@@ -320,60 +321,60 @@ def _process_fills(self, transaction: Any, in_transaction_list: List[InTransacti
...
+                    exchange=self.__COINBASE_PRO,
+                    holder=self.account_holder,
+                    transaction_type="Sell",
+                    spot_price=str(from_currency_price),
+                    crypto_out_no_fee=str(from_currency_size),
+                    crypto_fee="0",
+                    crypto_out_with_fee=None,
+                    fiat_out_no_fee=None,
+                    fiat_fee=None,
+                    notes=f"Sell side of conversion: {from_currency_size:.8f} {from_currency} -> {to_currency_size:.8f} {to_currency}",
```
houqp, line 338: best practice in python to use keyword args calling style to avoid passing in the wrong argument with the wrong index especially for functions that takes a long list of arguments.

src/dali/plugin/input/rest/coinbase_pro.py
eprbell: Nice, one liner fixes are my favorite :)

eprbell:
Thanks for fixing a bug!

PR #11, opened 2022-03-24
complete blockfi support

houqp:
Added support for parsing the following transaction types in all
transaction export CSV file:

* Trade
* Ach Deposit
* Ach Withdrawal
* Referral Bonus

Also added support for parsing all trades CSV export and merge with all
transactions exports.


src/dali/dali_main.py
```diff
@@ -290,6 +290,14 @@ def _validate_plugin_configuration(ini_config: ConfigParser, plugin_name: str, c
 
     for parameter in constructor_signature.parameters:
         annotation: Any = constructor_signature.parameters[parameter].annotation
+        if getattr(annotation, '__origin__', None) is Union \
+                and len(annotation.__args__) == 2 \
+                and annotation.__args__[1] is type(None):
+            if ini_config[plugin_name].get(parameter) is not None:
+                annotation = annotation.__args__[0]
+            else:
+                result[parameter] = None
```
houqp, line 299: adds support for optional plugin arguments.

src/dali/dali_main.py
```diff
@@ -290,6 +290,14 @@ def _validate_plugin_configuration(ini_config: ConfigParser, plugin_name: str, c
 
     for parameter in constructor_signature.parameters:
         annotation: Any = constructor_signature.parameters[parameter].annotation
+        if getattr(annotation, '__origin__', None) is Union \
+                and len(annotation.__args__) == 2 \
+                and annotation.__args__[1] is type(None):
```
eprbell, line 295: Is NoneType always guaranteed to be at position 1 with optional types?

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -50,11 +55,12 @@ def __init__(
         self,
         account_holder: str,
         csv_file: str,
```
eprbell, line 57: Rename 'csv_file' to 'transaction_csv_file', for clarity and simmetry with 'trade_csv_file'.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -76,7 +82,8 @@ def load(self) -> List[AbstractTransaction]:
                 if last_withdrawal_fee is not None and line[self.__TYPE_INDEX] != _WITHDRAWAL:
                     raise Exception(f"Internal error: withdrawal fee {last_withdrawal_fee} is not followed by withdrawal")
 
-                if line[self.__TYPE_INDEX] == _INTEREST_PAYMENT:
+                entry_type = line[self.__TYPE_INDEX]
```
eprbell, line 85: Can you add type hint: 
``` suggestion
entry_type: str = line[self.__TYPE_INDEX]
```

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -93,15 +100,34 @@ def load(self) -> List[AbstractTransaction]:
...
+                            transaction_type="Income",
+                            spot_price=Keyword.UNKNOWN.value,
+                            crypto_in=line[self.__AMOUNT_INDEX],
+                            fiat_fee="0",
+                            notes="Referral Bonus"
+                        )
+                    )
+                elif entry_type == _CRYPTO_TRANSFER:
+                    last_withdrawal_fee = None
+                    asset = line[self.__CURRENCY_INDEX]
```
eprbell, line 123: This is the only place where asset is assigned to a variable: for consistency with the rest of the code, can we remove the variable and just use `line[self.__CURRENCY_INDEX]` inline? An alternative would be to factor asset at the top of the file into a variable, but then we would have to do the same to all other "factorable" values (e.g. timestamp). I lean toward the first option.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -111,7 +137,25 @@ def load(self) -> List[AbstractTransaction]:
                             crypto_received=line[self.__AMOUNT_INDEX],
                         )
                     )
-                elif line[self.__TYPE_INDEX] == _WITHDRAWAL:
+                elif entry_type == _ACH_WITHDRAWAL:
```
eprbell, line 140: Are crypto sales marked as "Ach Withdrawal" in BlockFi? I was confused by this because ACH brings to mind a movement of fiat funds, but, based on this code, it looks like they mean sale of crypto (into fiat). I didn't do any trades on BlockFi myself, so don't have a trade CSV file to look at.

src/dali/dali_main.py
houqp: yes based from my tests.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -111,7 +137,25 @@ def load(self) -> List[AbstractTransaction]:
...
+                        OutTransaction(
+                            plugin=self.__BLOCKFI,
+                            unique_id=Keyword.UNKNOWN.value,
+                            raw_data=raw_data,
+                            timestamp=f"{line[self.__TIMESTAMP_INDEX]} -00:00",
+                            asset=line[self.__CURRENCY_INDEX],
+                            exchange=self.__BLOCKFI,
+                            holder=self.account_holder,
+                            transaction_type="Sell",
+                            spot_price=Keyword.UNKNOWN.value,
```
eprbell, line 152: Do they have spot price in the trade file? They used to have it in the transaction file up until last year, but then they removed it recently. I asked their customer support if they could put it back, and they basically said: "No"... 

Hopefully they'll release a REST API at some point (they have been saying it's in their roadmap for at least one year).

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -111,7 +137,25 @@ def load(self) -> List[AbstractTransaction]:
...
+                            crypto_out_no_fee=str(-RP2Decimal(line[self.__AMOUNT_INDEX])),
+                            crypto_fee="0",
+                            notes="ACH redrawal",
```
eprbell, line 155: It's "ACH withdrawal". But again, I think referral to ACH is mostly confusing to the final user: remove the note altogether?

src/dali/plugin/input/csv/blockfi.py
houqp: USD (fiat) redrawals in blockfi are tracked as GUSD tokens

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
                             crypto_received=Keyword.UNKNOWN.value,
                         )
                     )
-                elif line[self.__TYPE_INDEX] == _WITHDRAWAL_FEE:
+                elif entry_type == _ACH_DEPOSIT:
```
eprbell, line 180: Same Ach comment as above.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+                            raw_data=raw_data,
+                            timestamp=f"{line[self.__TIMESTAMP_INDEX]} -00:00",
+                            asset=line[self.__CURRENCY_INDEX],
+                            exchange=self.__BLOCKFI,
+                            holder=self.account_holder,
+                            transaction_type="Buy",
+                            spot_price=Keyword.UNKNOWN.value,
+                            crypto_in=line[self.__AMOUNT_INDEX],
+                            fiat_fee="0",
+                            notes="Ach deposit",
```
eprbell, line 195: Same as above: remove the notes assignment?

houqp:
> I think referral to ACH is mostly confusing to the final user: remove the notes assignment altogether?

IMHO, this makes it less confusing. It's mostly a UX issue that's introduced from blockfi because they model all fiat transactions as GUSD tokens. Without this annotation, it's not possible to tell whether the transaction was created from a token swap or a fiat deposit/withdrawal.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+                elif entry_type == _TRADE:
+                    # Trades will be handled by parsing trade_report_all.csv
+                    # export
+                    continue
+                elif entry_type == _WITHDRAWAL_FEE:
                     last_withdrawal_fee = RP2Decimal(line[self.__AMOUNT_INDEX])
                     last_withdrawal_fee = -last_withdrawal_fee  # type: ignore
                 else:
-                    self.__logger.debug("Unsupported transaction type (skipping): %s", raw_data)
+                    self.__logger.error("Unsupported transaction type (skipping): %s", raw_data)
```
eprbell, line 205: Nice.

houqp:
> On the other hand if the transaction refers to fiat movements we can just remove the entire elif snippet, as discussed above.

Could you elaborate on this? Are you suggesting that we skipping generating a transaction for fiat movements?

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+            result += self.parse_trade_report(self.__trade_csv_file)
+        return result
+
+    def parse_trade_report(self, file_path: str) -> List[AbstractTransaction]:
+        result: List[AbstractTransaction] = []
+
+        with open(file_path, mode="r", encoding="utf-8") as csv_file:
+            lines = reader(csv_file)
+            # skip csv header
+            header = next(lines)
```
eprbell, line 217: This is good: would you mind doing this in the other csv loop as well? I was using a more cumbersome method.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+
+            col_idx = {}
+            for (idx, name) in enumerate(header):
+                col_idx[name] = idx
+
+            for line in lines:
+                raw_data: str = self.__DELIMITER.join(line)
+                self.__logger.debug("Transaction: %s", raw_data)
+
+                trans_type = line[col_idx["Type"]]
```
eprbell, line 228: Could you use constants rather than strings?

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+                if trans_type != "Trade":
```
eprbell, line 229: The code uses full words in identifiers:
```suggestion
                 if transaction_type != "Trade":
```

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+                        unique_id=f"{trade_id}/buy",
+                        raw_data=raw_data,
+                        timestamp=timestamp,
+                        asset=to_currency,
+                        exchange=self.__BLOCKFI,
+                        holder=self.account_holder,
+                        transaction_type="Buy",
+                        spot_price=Keyword.UNKNOWN.value,
+                        crypto_in=to_size,
+                        fiat_fee="0",
```
eprbell, line 268: This looks like a crypto -> crypto conversion: in this case doesn't BlockFi charge a crypto fee? If so it can be captured by using the crypto_fee of InTransaction (introduced very recently in both DaLI and RP2): see CoinbasePro coin swap for an example.

houqp:
> This is good: would you mind doing this in the other csv loop as well? I was using a more cumbersome method.

I recommend doing this refactoring in a separate PR so the commits are logically separated.

src/dali/plugin/input/csv/blockfi.py
eprbell, line 202: If this is a crypto buy (not a fiat movement) I would remove the notes assignment, because I imagine "ACH" would be confusing to the final user.

houqp: ACH deposit is a fiat movement, not a crypto buy, it's just that blockfi tracks it as GUSD token.

src/dali/plugin/input/csv/blockfi.py
houqp: blockfi doesn't charge crypto fees

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+    def parse_trade_report(self, file_path: str) -> List[AbstractTransaction]:
+        result: List[AbstractTransaction] = []
+
+        with open(file_path, mode="r", encoding="utf-8") as csv_file:
+            lines = reader(csv_file)
+            # skip csv header
+            header = next(lines)
+            self.__logger.debug("Header: %s", header)
+
+            col_idx = {}
```
eprbell, line 228: Can you add type hints?

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -133,9 +177,98 @@ def load(self) -> List[AbstractTransaction]:
...
+                raw_data: str = self.__DELIMITER.join(line)
+                self.__logger.debug("Transaction: %s", raw_data)
+
+                trans_type = line[col_idx["Type"]]
+                if trans_type != "Trade":
+                    raise Exception(f"Internal error: unsupported transaction type: {trans_type}")
+
+                trade_id = line[col_idx["Trade ID"]]
+                date = line[col_idx["Date"]]
+                timestamp = f"{date} -00:00"
```
eprbell, line 242: Type hints

eprbell (changes requested):
I have a few comments here and there, but overall this looks great!

src/dali/plugin/input/csv/blockfi.py
houqp: BTW, I think a lot of the type hints in this project are not needed because mypy does type inference like most modern programming languages. Adding unnecessary type hints does make the code less readable compared to other idiomatic python projects. I am usually the guy who asks people to add type annotations in python code base during code reviews, but I think we are over doing it in rp2 :P

houqp: For these couple lines, mypy should have the info to make the inference that these variables are all strings.

src/dali/plugin/input/csv/blockfi.py
houqp: added

src/dali/plugin/input/csv/blockfi.py
eprbell: Ok, then this is fine. GUSD has slight price variations compared to USD and so it can trigger taxable events, I guess, so it's ok to track it. I'm asking these questions because my BlockFi activity is fairly simple and I didn't see the transaction types you introduced with this PR.

src/dali/plugin/input/csv/blockfi.py
eprbell: Ok.

src/dali/plugin/input/csv/blockfi.py
eprbell: I would disagree on the code being less readable with type hints: yes, mypy does type inference but type hints make type declaration explicit (vs implicit with inference), which to me makes the code clearer. Also by adding types everywhere we remove the arbitrary decision of when to add types vs not. But I don't want to make this into a big thing for such a small snippet: it's ok to leave them out :-)

src/dali/plugin/input/csv/blockfi.py
eprbell: Oh, really? I wish Coinbase did the same :-D (they charge handsomely).

src/dali/dali_main.py
eprbell: Ok.

src/dali/dali_main.py
eprbell: Cool.

eprbell (approved):
Good stuff, thanks!

src/dali/plugin/input/csv/blockfi.py
houqp: blockfi hides the fee through higher premiums :face_exhaling: at least coinbase is being transparent :)

src/dali/plugin/input/csv/blockfi.py
houqp: yeah, unfortunately it results in unnecessary gains. blockfi also sends out 1099 to IRS with each transaction tracked in GUSD, so there is no easy way to work around it.

houqp:
Sorry should I have converted this PR to draft mode before going to bed, was planning to squash the commits to keep the history clean.

src/dali/plugin/input/csv/blockfi.py
houqp: > Also by adding types everywhere we remove the arbitrary decision of when to add types vs not. 

I agree this is a useful rule to establish to make code reviews more efficient for both sides. The rule I usually apply is to make sure function/method signatures are annotated and only annotate the rest of the code base enough to pass mypy. Occasionally, I will annotate a variable if it really helps readability, but usually well written code is self-explanatory and so this doesn't happen often. Type annotations are mostly for compilers, humans usually can figure the types out with good variable/function names plus context. For example when I read a variable named `transaction_list`, I would implicitly assume it to be a list of transactions. So to me, `transaction_list: List[AbstractTransaction]` makes the code more verbose and hard to read. Same for `amount: RP2Decimal = RP2Decimal(transaction[_AMOUNT][_AMOUNT])`, the type for the amount variable should be obvious for a human and a machine. IMHO, redundant info hurts readability. It's certainly not an easy balance to strike because lack of essential info also hurts readability.

I assume you work with a large monitor and editor window. For smaller windowses, excessive type annotations plus long variable names make things worse because many times I won't be able to see the full line without performing horizontal scrolling in my editor. Human eyes are better at scanning lines vertically instead of horizontally. Adding horizontal window scroll makes it even worse. Here is a screenshot of what it looks like in my coding environment:

vim-screen-shot

This is why the official python style guide recommends a max line length of 79 chars: https://peps.python.org/pep-0008/#maximum-line-length. I try to not be doctrinal about this because there are rare cases where going pass that 79 limit improves readability, but I think in general, it's a good practice to keep lines short.

Something is not right if the python code is a lot more verbose than the Go and Rust code I write these days ;)

eprbell: I have used strongly typed languages for a long time and am of the school of thought that explicit types are a good thing: it's a zero-tolerance approach against potential ambiguity, even when the ambiguity is little or none (again, removing the decision is the main point). I have lived through many horror stories, where compiler warnings were disabled and inexplicable bugs occurred due to uncaught typing errors (but in fairness that was with C, not Python, so it's a bit apples/oranges). In this case I guess delegating the decision to mypy is an acceptable compromise.

As for monitors, I do have a large monitor and imagine many coders do (the price of monitors has been falling for many years, reducing the barrier of entry to getting a large one). I always thought the 80 character limit thing was a legacy from the 80s that never died, but I admit I like your point about human eyes being better at scanning vertically. :)

PR #14, opened 2022-03-25
record fee for both in/out transactions in crypto swap

houqp:
@eprbell I don't have any sell side fills in my transaction log to validate whether this logic is correct or not. but for buy side fills, this PR fixes the following when I swap BTC for ETH:

* coinbase pro deducts the transaction fee from BTC, so crypto fee should be recorded in the out transaction
* in order to add the transaction fee ETH's cost basis, we need to record the fee as fiat amount in in transaction

As a result, the fee won't be considered as part of the proceeds when calculating gain/loss for the BTC out transaction, and it will be added into the cost basis for the ETH in transaction so when we sell ETH in the future, we get a lower cost basis.

Without this change, a bunch of ETH out transactions will be created for the recorded crypto fees, which incorrectly reduces my ETH balance.


houqp:
@eprbell do we have existing tests for coinbase pro plugin so I can write some tests for both buy and sell fill transactions?

src/dali/plugin/input/rest/coinbase_pro.py
```diff
@@ -334,8 +334,8 @@ def _process_fills(self, transaction: Any, in_transaction_list: List[InTransacti
                     transaction_type="Sell",
                     spot_price=str(from_currency_price),
                     crypto_out_no_fee=str(from_currency_size),
-                    crypto_fee="0",
-                    crypto_out_with_fee=None,
+                    crypto_fee=str(crypto_fee),
+                    crypto_out_with_fee=str(from_currency_size+crypto_fee),
```
eprbell, line 338: You can leave `crypto_out_with_fee=None`, because RP2 computes it automatically as` crypto_out_no_fee + crypto fee` (see https://github.com/eprbell/rp2/blob/main/src/rp2/out_transaction.py#L69).

src/dali/plugin/input/rest/coinbase_pro.py
```diff
@@ -354,10 +354,10 @@ def _process_fills(self, transaction: Any, in_transaction_list: List[InTransacti
                     transaction_type="Buy",
                     spot_price=str(to_currency_price),
                     crypto_in=str(to_currency_size),
-                    crypto_fee=str(crypto_fee),
-                    fiat_in_no_fee=None,
+                    crypto_fee=None,
+                    fiat_in_no_fee=str(usd_volume),
                     fiat_in_with_fee=None,
-                    fiat_fee=None,
+                    fiat_fee=str(from_currency_price * crypto_fee),
```
eprbell, line 360: A couple of questions (I'm shooting in the dark, because I don't have a buy-side fill log to look at):
- is fiat_in_no_fee == crypto_in * spot_price)? If it is you can leave fiat_in_no_fee=None and RP2 will compute it automatically (https://github.com/eprbell/rp2/blob/main/src/rp2/in_transaction.py#L77).
- also a bit curious why you need to set the fiat_fee, because it looks to me like you're paying the fee twice (once as crypto_fee in the OutTransaction and once here in the InTransaction).

eprbell:
Interesting: I did the opposite swap (I swapped ETH for BTC) and so I only have the sell-side: this is good, because it means we should be able to come up with the right behavior by analyzing your situation and mine. The existing code handles correctly the sell-side.

I have a unit test but it's based on my personal information so I can't publish it on Github: I just run it locally on my machine before pushing code (it uses tests/ods_diff.py to compare the output against a golden file). Is there a way to test the Coinbase Pro plugin without exposing your info?

houqp:
> I just run it locally on my machine before pushing code (it uses tests/ods_diff.py to compare the output against a golden file). Is there a way to test the Coinbase Pro plugin without exposing your info?

Yeah, I can convert my transactions to unit tests.

src/dali/plugin/input/rest/coinbase_pro.py
houqp: > also a bit curious why you need to set the fiat_fee, because it looks to me like you're paying the fee twice (once as crypto_fee in the OutTransaction and once here in the InTransaction).

The fee is only paid once, I am setting the fee in in transaction so that the fee will be taken into account when calculating the cost basis in future sales. Technically if we set `crypto_fee="0"` for the out transaction, we will still get the same tax calculation. But since in reality the transaction fee was paid using the from_currency (btc in my example) in coinbase pro, i figured it would be more accurate to record it as crypto fee as part of the out transaction.

eprbell: When you set crypto_fee in the OutTransaction, RP2 will automatically convert it to fiat and populate fiat_fee (https://github.com/eprbell/rp2/blob/main/src/rp2/in_transaction.py#L66). The fiat_fee is taken into account and used to lower proceeds for the OutTransaction (see https://github.com/eprbell/rp2/issues/6): you get the tax benefit in this way. This is why I'm thinking there should be no need to set fiat_fee in the InTransaction and to raise the cost basis (because the fee was already used to lower the proceeds in the OutTransaction). I think this code might both lower your proceeds in the OutTransaction and raise the cost basis in the InTransaction, thus giving a double deduction.

eprbell: So after some more thinking over this, I think the correct implementation would be:
* line 315 (inside `if fill_side == _SELL`), add two lines:
```
out_crypto_fee: RP2Decimal = ZERO
in_crypto_fee: RP2Decimal =  crypto_fee
```
* line 321 (inside `if fill_side == _BUY`), add two lines:
```
out_crypto_fee: RP2Decimal = crypto_fee
in_crypto_fee: RP2Decimal =  ZERO
```
And then assign `out_crypto_fee` and `in_crypto_fee` to the crypto_fee field of the Out/InTransaction. There is no need to modify fiat_fee.

houqp: > I think this code might both lower your proceeds in the OutTransaction and raise the cost basis in the InTransaction, thus giving a double deduction.

This was my intention as noted in my PR description :) Since this is not well defined in the IRS code, I figured we could count it as fees for both the buy side and sell side.

But I agree with you that it it would be safer to only count the fee on one side of the trade. Will update the code with a test case for buy side.

setup.cfg
```diff
@@ -37,6 +37,25 @@ install_requires =
     rp2>=0.9.25
     yfinance>=0.1.70
 
+[options.extras_require]
+dev =
```
houqp, line 41: managing dev dependency using extras is more idiomatic across python projects. we also don't need to repeat the install_requires dependencies in the requirements.txt file. In fact, I think we can just get rid of the requirements.txt file and use `pip install '.[dev]'` everywhere :)

setup.cfg
```diff
@@ -37,6 +37,25 @@ install_requires =
...
+    pytest
```
eprbell, line 42: Pytest already appears below. Also can you put them in alphabetic order?

setup.cfg
eprbell: Didn't know this, sounds good. Could you update also the Makefile (which is using requirements.txt)?

tests/test_plugin_coinbase_pro.py
```diff
@@ -0,0 +1,118 @@
+from typing import List, Dict, Any
```
eprbell, line 1: This is really cool: I hadn't used pytest-mock before. A couple of minor nits:
- can you add a copyright and license header similar to the one at the beginning of all other source files (e.g. tests/test_ods_output_diff.py)
- can you use Python unit tests (they are used both in DaLI and RP2). You don't have to do it, if you don't have time: I can migrate it later.

eprbell (changes requested):
Looks good and I love the unit test with pytest-mock: I had been wondering how to test REST plugins and now I know :-D. Only a couple of minor nits and we're good to go.

tests/test_plugin_coinbase_pro.py
houqp: For copyright license header, i recommend using tools like rat to enforce it during code review like other ASF projects if you feel strongly about it.

I changed the test function to a class to better align with the structure you have with tests using uniitest module. But honestly, I recommend stop using the unittest module going forward because it doesn't add much value. Pytest is designed to work with functions and fixtures. I tend to find tests declared in functions easier to read and reason because code is less nested and dependencies are clearly declared in the function signatures. With a class style enforced by the unittest module, it's never clear how much states are shared between tests without scrolling back to the class init method to see what class attributes are defined. I have switched to pure function tests in all my python projects for many years and it scales very well for small to very large projects.

The biggest problem with the unnittest module is it doesn't play well with pytest's fixture feature, that's why it's recommended to not use it for pytest tests. See the first note at the bottom of https://pytest.org/en/latest/how-to/unittest.html. The unittest support was added to pytest only for backwards compatibility reasons.

tests/test_plugin_coinbase_pro.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2022 eprbell
```
eprbell, line 1: This file was created by you, so this should be your copyright.

tests/test_plugin_coinbase_pro.py
eprbell: Thanks for the suggestions. I will definitely look into rat. As for unit tests, you make a compelling case: going forward, let's move away from unit test, or maybe allow both, so that it's up to the test creator to decide what to use.

eprbell (approved):

tests/test_plugin_coinbase_pro.py
houqp: yeah, it's very typical to allow both as well, especially for projects with longer histories :)

eprbell (approved):

PR #17, opened 2022-03-26
track transfers across blockfi and coinbase wallets

houqp:
blockfi reports cost basis as 0 for transfers, this PR fixes it.

### DIFF
--- a/src/dali/dali_configuration.py
+++ b/src/dali/dali_configuration.py
@@ -60,6 +60,7 @@ class Keyword(Enum):
     TIMESTAMP: str = "timestamp"
     TO_EXCHANGE: str = "to_exchange"
     TO_HOLDER: str = "to_holder"
+    TO_ADDRESS: str = "to_address"
     TRANSACTION_HINTS: str = "transaction_hints"
     TRANSACTION_TYPE: str = "transaction_type"
     UNIQUE_ID: str = "unique_id"
@@ -102,6 +103,7 @@ _INTERNAL_FIELD_SET: Set[str] = {
     Keyword.IS_SPOT_PRICE_FROM_WEB.value,
     Keyword.PLUGIN.value,
     Keyword.RAW_DATA.value,
+    Keyword.TO_ADDRESS.value,
 }
 
 DIRECTION_SET: Set[str] = {
@@ -178,6 +180,7 @@ DEFAULT_CONFIGURATION: Dict[str, Dict[str, int]] = {
         Keyword.CRYPTO_RECEIVED.value: 8,
         Keyword.UNIQUE_ID.value: 12,
         Keyword.NOTES.value: 13,
+        Keyword.TO_ADDRESS.value: 14,
     },
 }
 
--- a/src/dali/plugin/input/csv/blockfi.py
+++ b/src/dali/plugin/input/csv/blockfi.py
@@ -63,11 +63,17 @@ class InputPlugin(AbstractInputPlugin):
         account_holder: str,
         transaction_csv_file: str,
         trade_csv_file: Optional[str] = None,
+        addresses: Optional[str] = None,
     ) -> None:
 
         super().__init__(account_holder)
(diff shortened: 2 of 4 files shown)


houqp:
tests are failing due to https://github.com/eprbell/rp2/pull/20

eprbell:
I haven't looked in detail at this yet, but I don't think it's necessary to modify RP2. The new field (to_address) can live only in DaLI and DaLI doesn't even need to generate it in the output (this will also make the tests pass).

eprbell:
To avoid generating `to_address`, just add it to _INTERNAL_FIELD_SET in dali_configuration.py.

src/dali/dali_configuration.py
```diff
@@ -178,6 +180,7 @@ def type_check_from_string(cls, keyword: str) -> "Keyword":
         Keyword.CRYPTO_RECEIVED.value: 8,
         Keyword.UNIQUE_ID.value: 12,
         Keyword.NOTES.value: 13,
+        Keyword.TO_ADDRESS.value: 14,
```
eprbell, line 183: Remove this: it's for generated fields.

src/dali/transaction_resolver.py
```diff
@@ -151,7 +166,34 @@ def resolve_transactions(
         if not isinstance(transaction, AbstractTransaction):
             raise Exception(f"Internal error: Parameter 'transaction' is not a subclass of AbstractTransaction. {transaction}")
 
+        if isinstance(transaction, IntraTransaction) and is_unknown(transaction.to_exchange):
+            # transfer out to unknown destination, record out address and
+            # amount for matching
+            if transaction.from_exchange == _COINBASE:
+                raw_data = json.loads(transaction.raw_data)
+                to_address = raw_data[_TO][_ADDRESS]
```
eprbell, line 174: This breaks abstraction layers: the transaction resolver should not have plugin-specific code.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -63,11 +63,17 @@ def __init__(
         account_holder: str,
         transaction_csv_file: str,
         trade_csv_file: Optional[str] = None,
+        addresses: Optional[str] = None,
```
eprbell, line 66: This doesn't seem the right model: many addresses could point to the same account.

eprbell:
This is an interesting attempt, but I think it doesn't work as we expect: it breaks abstraction layers (the transaction resolver shouldn't have plugin-specific code) and the single-address-per-currency model doesn't reflect crypto semantics (multiple addresses can point to the same wallet). 

I think the problem of BlockFi not being fully integrated is valid though, and there are a few of things we can try to approach the issue:
1. BlockFi has repeatedly said they have a REST API on their roadmap (https://twitter.com/BlockFi/status/1507433260380524557):  it's not clear when it will happen, but this would be the cleanest solution;
2. heuristics-based approach: a while ago (before uploading the first version on Github) I came up with some heuristics (activated with CLI option) that tried to match "unknown" transactions based on "closeness" of timestamp and amount. It wasn't bulletproof but worked reasonably well with my blockfi transactions. In the end I discarded it exactly because it wasn't bulletproof and I preferred approach 3 (see below). Perhaps we can attempt this once again (I'm not sure if I kept the code, but I'll look);
3. manual CSV: this is already doable today (it's what I do with my BlockFi taxes). Generate a BlockFi-only output with DaLI and then convert it to a manual plugin format and fill the missing information, then feed the manual CSV to DaLI with the rest of the transactions. Not ideal, but it works.

My strong preference would be for 1. Absent that, I'd go for 3. I'm ok with experimenting with 2, as long as it's disabled by default and users can activate with CLI option (e.g. --experimental-heuristic-resolution), so that they know it may generate bad results.

eprbell:
Closing this for now. Please reopen if you want to attempt one the approaches mentioned above.

Issue #15, opened 2022-03-26
Coinbase REST API: Conversions - Converted From Appear as Negative Value "Buy" in .ods Output file


jameskupke:
Version: 0.3.21
OS: Fedora Linux 35

On Reviewing transactions outputted from DaLI from the coinbase API, conversions do not appear to be entered correctly.

For example I have a conversion of $2.97 worth of ALGO to BTC, there is a IN entry of BUY -$2.97 in the ALGO tab.

eprbell:
First of all, just wanted to make sure: this is Coinbase, not Coinbase Pro, correct? The Coinbase plugin doesn't support conversions yet (because I don't have an example of them in my transactions). 

Would you be willing to share the one line from your debug logs that shows the JSON returned by Coinbase for this transaction? Before I explain how to do that, let me point out that if you're not comfortable sharing this information, you have a couple of additional options to consider:
* remove personal information from the line: in particular timestamps and ids. If you want to remove/mangle amounts you can do itm but that will probably make it harder to come up with a correct implementation (if you do that, be sure to preserve sign information);
* if you don't want to share the line publicly you also have the option of sharing it privately with me (my Twitter DMs are open: @eprbell).

With that out of the way here's how to do it:
* prepend your dali command line with `LOG_LEVEL=DEBUG `
* after running dali in debug mode, cd to the log directory, open the latest log and grep for the line of the culprit transaction (check in the Coinbase section of the log)

jameskupke:
This is Coinbase, not Coinbase Pro.

I believe here's a transaction:
I masked the date a bit and masked the ids; there appears to be 3 different ids, so I marked with X,Y,Z.


```
2022-03-26 08:27:35,006/dali/DEBUG: Self-contained transaction: InTransaction:
  plugin=Coinbase
  unique_id=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
[...]

PR #18, opened 2022-03-26
support ETH/ETH2 swaps in coinbase

houqp:
without this fix, it records a ETH buy transaction with a negative amount

### DIFF
--- a/src/dali/plugin/input/rest/coinbase.py
+++ b/src/dali/plugin/input/rest/coinbase.py
@@ -374,14 +374,16 @@ class InputPlugin(AbstractInputPlugin):
             return
 
         raw_data: str = json.dumps(transaction)
-        if transaction_type in {_BUY, _TRADE}:
-            spot_price = (native_amount - fiat_fee) / crypto_amount
+        if transaction_type == _BUY or (transaction_type == _TRADE and native_amount >= ZERO):
             if transaction_type == _BUY:
                 buy: Dict[str, Any] = id_2_buy[transaction[transaction_type][_ID]]
                 raw_data = f"{raw_data}//{json.dumps(buy)}"
                 fiat_fee = RP2Decimal(buy[_FEE][_AMOUNT])
                 self.__logger.debug("Buy: %s", json.dumps(buy))
                 spot_price = RP2Decimal(buy[_UNIT_PRICE][_AMOUNT])
+            else:
+                # swap in transaction
+                spot_price = (native_amount - fiat_fee) / crypto_amount
             in_transaction_list.append(
                 InTransaction(
                     plugin=self.__COINBASE,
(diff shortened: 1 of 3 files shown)


eprbell:
This will fix issue #15. Reviewing now...

houqp:
I will be adding a test to guard against regression.

src/dali/plugin/input/rest/coinbase.py
```diff
@@ -390,7 +396,7 @@ def _process_fill(
                 )
             )
         else:
-            self.__logger.debug("Unsupported transaction type (skipping): %s", json.dumps(transaction_type))
+            self.__logger.warning("Unsupported transaction type (skipping): %s", transaction_type)
```
eprbell, line 399: Should we do this (like elsewhere in the code)?
```suggestion
            self.__logger.error("Unsupported transaction type (skipping): %s. Please open an issue at %s", raw_data, self.ISSUES_URL)

```

src/dali/plugin/input/rest/coinbase.py
```diff
@@ -339,13 +339,15 @@ def _process_fill(
...
-            spot_price = (native_amount - fiat_fee) / crypto_amount
+        if transaction_type == _BUY or (transaction_type == _TRADE and native_amount >= ZERO):
             if transaction_type == _BUY:
                 buy: Dict[str, Any] = id_2_buy[transaction[transaction_type][_ID]]
                 fiat_fee = RP2Decimal(buy[_FEE][_AMOUNT])
                 self.__logger.debug("Buy: %s", json.dumps(buy))
                 spot_price = RP2Decimal(buy[_UNIT_PRICE][_AMOUNT])
+            else:
+                # swap in transaction
+                spot_price = (native_amount - fiat_fee) / crypto_amount
```
eprbell, line 350: In this path fiat_fee is always 0: this may be the correct for your specific case (is Coinbase not charging fee when swapping ETH -> ETH2?), but I suspect when swapping coins there will be a fee and it will be probably paid in crypto (similar to what happens on Coinbase Pro). Do you have another example of coin swap handy other than ETH -> ETH2? If not we can follow-up on issue #15: @jameskupke has posted the Buy half of an Algo -> XTZ conversion. We need to get the Sell side as well. 

Note that in RP2 passing a crypto_fee to an InTransaction is not the equivalent to passing a fiat_fee. If fiat_fee is assigned, crypto_fee is set to 0 by RP2. If crypto_fee is assigned, fiat_fee is set to crypto_fee * spot_price by RP2. The reason for this behavior is that if the fee is paid in fiat, then no crypto is used for the fee, but if the fee is paid in crypto, then its converted fiat value is needed to compute taxes. Note that RP2 models a non-zero crypto_fee with a separate fee-typed out-transaction.

In your case was there a fee and if so was it crypto or fiat?

eprbell: Also if there was a crypto fee was it paid in ETH or ETH2?

eprbell: To summarize my position: I think this PR is good to start with (it does solve the negative amount problem), but down the road we may have to modify the _TRADE branches to account for crypto fees.

houqp: There is no fee based on the api response when you stake ETH for ETH2. The PR only focuses on ETH2 stake because that's the only swap transaction I have in coinbase :) I am pretty sure more work is needed to handle other type of swaps.

eprbell: Sounds good, let's go with what we have and we'll keep an eye out for crypto fees later.

eprbell (approved):

eprbell (approved):

PR #31, opened 2022-04-15
Coinbase historical data API expects UTC timestamps only.

stevendavis:
This is a small code change to force UTC timestamps when requesting historical data from Coinbase via the Historic_Crypto package.  The Coinbase API expects the from time stamp and to timestamp will be UTC timezone.

Unfortunately, the TestODSOutputDiff unit tests are failing. I may be mistaken, but I think the unit test expected results are incorrect and will need to be updated.


eprbell:
Do you have a pointer to the documentation mentioning that timezone should be UTC?

stevendavis:
The API is not thoroughly documented, but there is a discussion of timestamps and timezones on the forum here: https://forums.coinbasecloud.dev/t/invalid-end-on-product-candles-endpoint/320

src/dali/transaction_resolver.py
```diff
@@ -127,12 +127,13 @@ def _update_spot_price_from_web(transaction: AbstractTransaction, historical_pri
             LOGGER.debug("Reading spot_price for %s/%s from cache: %s", key.timestamp, key.asset, spot_price)
         else:
             time_granularity: List[int] = [60, 300, 900, 3600, 21600, 84600]
-            from_timestamp: str = transaction.timestamp_value.strftime("%Y-%m-%d-%H-%M")
+            transaction_utc_timestamp = transaction.timestamp_value.astimezone(timezone.utc)  # Coinbase API expects UTC timestamps only
```
eprbell, line 130: Could you add the link you found (https://forums.coinbasecloud.dev/t/invalid-end-on-product-candles-endpoint/320) on this in the comment?

eprbell (changes requested):
Thanks for the reasearch, the link and the bug fix: good work! The ODS tests are failing because now the spot price for some transactions has changed. To fix that cp the failing files from output/ to input/golden.

eprbell:
This looks good. Just update the files in input/golden and I'll approve: 
```
cp output/test_ods_output_diff/*.ods input/golden/
```

eprbell (approved):
Ship it!

PR #37, opened 2022-04-17
Allow epoch timestamp

macanudo527:
For Binance.com (and possibly other exchanges) the rest API returns epoch time in milliseconds. It seems a little silly to format the epoch time only to be parsed later. Is it possible to check if the timestamp is an epoch timestamp and if so, just use that to create the timestamp_value datetime variable.

### DIFF
--- a/src/dali/abstract_transaction.py
+++ b/src/dali/abstract_transaction.py
@@ -73,7 +73,10 @@ class AbstractTransaction:
     def _validate_timestamp_field(cls, name: str, value: str, raw_data: str) -> StringAndDatetime:
         value = cls._validate_string_field(name, value, raw_data, disallow_empty=True, disallow_unknown=True)
         try:
-            result: datetime = parse(value)
+            if value.isdigit():
+                result: datetime = datetime.fromtimestamp(value)
+            else:
+                result: datetime = parse(value)
         except Exception as exc:
             raise Exception(f"Internal error parsing {name} as datetime: {value}\n{raw_data}\n{str(exc)}") from exc
         if result.tzinfo is None:


src/dali/abstract_transaction.py
```diff
@@ -73,7 +73,10 @@ def _validate_numeric_field(cls, name: str, value: str, raw_data: str, disallow_
     def _validate_timestamp_field(cls, name: str, value: str, raw_data: str) -> StringAndDatetime:
         value = cls._validate_string_field(name, value, raw_data, disallow_empty=True, disallow_unknown=True)
         try:
-            result: datetime = parse(value)
+            if value.isdigit():
```
eprbell, line 76: I'm a little hesitant in adding this logic to abstract_transaction: being a core class I'd prefer it having an unambiguous API contract (e.g. the timestamp is always expressed as a date string, never as an epoch) that is easy to validate. The design principle is that the core API should not try to accomodate exchange peculiarities: it should be the other way around. So I think it would be best to do this conversion in the plugin: you can add a utility function that does the conversion within the plugin and is called as needed.

macanudo527: Ok, fair enough. In my mind, I guess I see the 'base' form of datetime as an epoch timestamp, but I guess it can really be anything. Thanks for clarifying. I'll go with your suggestion.

eprbell: Agree that epoch is more "base" but the two are essentially equivalent. My main concern was with not allowing both forms to be assignable to the same parameter, because that makes the API contract a little more ambiguous.

PR #51, opened 2022-06-01
Do not concatenate identical notes when resolving intra transactions.

stevendavis:
Added 1 line of code to transaction resolver.
Added 10 new test cases.

Are we all using the same version of isort?


tests/test_transaction_resolver.py
```diff
@@ -0,0 +1,80 @@
...
+    transaction1 = IntraTransaction(
+        plugin="plugin",
+        unique_id="unique_id",
+        raw_data="raw_data1",
+        asset="asset",
+        timestamp="2022-01-01 00:00:00+00:00",
+        from_exchange="from_exchange",
+        from_holder="from_holder",
+        to_exchange="to_exchange",
+        to_holder="to_holder",
```
eprbell, line 56: To make the test more realistic, make to_ fields unknown:
```suggestion
        to_exchange=Keyword.UNKNOWN.value,
        to_holder=Keyword.UNKNOWN.value,
```

tests/test_transaction_resolver.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+    transaction2 = IntraTransaction(
+        plugin="plugin",
+        unique_id="unique_id",
+        raw_data="raw_data2",
+        asset="asset",
+        timestamp="2022-01-01 00:00:00+00:00",
+        from_exchange="from_exchange",
+        from_holder="from_holder",
```
eprbell, line 70: To make the test more realistic, make from_ fields unknown:
```suggestion
        from_exchange=Keyword.UNKNOWN.value,
        from_holder=Keyword.UNKNOWN.value,
```

tests/test_transaction_resolver.py
```diff
@@ -0,0 +1,80 @@
...
+        to_exchange="to_exchange",
+        to_holder="to_holder",
+        spot_price=None,
```
eprbell, line 73: Assign `spot_price` (say, 1000 or something) to avoid network calls in the unit test. Leave the `spot_price` in transaction1 `None`.

tests/test_transaction_resolver.py
```diff
@@ -0,0 +1,80 @@
...
+        crypto_sent="1.0",
+        crypto_received="1.0",
+        notes=notes2,
+    )
+
+    resolved = _resolve_intra_intra_transaction(transaction1, transaction2, prior_notes)
+    assert resolved.notes == resolved_notes
```
eprbell, line 80: Add assertions that `from_exchange`, `from_holder`, `to_exchange`, `to_holder` contain the expected values. Add assertion that `spot_price` is what you expect it to be (1000 or whatever you assigned to it.

eprbell (changes requested):
LGTM (there is a small conflict to fix before merging). I like the transaction resolver test: there are many execution paths in the resolver that would benefit from unit tests, so it would be great to add more unit tests there. As for isort, it's specified in setup.cfg without a version constraint, so it should install the latest version when you setup the development workflow.

stevendavis:
What's going on with the pylint failures? 

************* Module dali.abstract_input_plugin
src/dali/abstract_input_plugin.py:36:0: E0012: Bad option value for disable. Don't recognize message no-self-use. (bad-option-value)
************* Module dali.plugin.pair_converter.historic_crypto
src/dali/plugin/pair_converter/historic_crypto.py:28:0: E0012: Bad option value for disable. Don't recognize message no-self-use. (bad-option-value)
************* Module test_plugin_coinbase
tests/test_plugin_coinbase.py:26:0: E0012: Bad option value for disable. Don't recognize message no-self-use. (bad-option-value)
************* Module test_plugin_coinbase_pro
tests/test_plugin_coinbase_pro.py:24:0: E0012: Bad option value for disable. Don't recognize message no-self-use. (bad-option-value)
************* Module test_plugin_historic_crypto
tests/test_plugin_historic_crypto.py:39:0: E0012: Bad option value for disable. Don't recognize message no-self-use. (bad-option-value)
tests/test_plugin_historic_crypto.py:94:0: E0012: Bad option value for disable. Don't recognize message no-self-use. (bad-option-value)

src/dali/transaction_resolver.py
```diff
@@ -125,6 +122,7 @@ def _get_originating_exchange(transaction: AbstractTransaction) -> str:
         return transaction.from_exchange
     raise Exception(f"Internal error: not a transaction: {transaction}")
 
+
```
eprbell, line 125: Remove extra newline.

eprbell:
> 

Strange... I see they [released a new version](https://pypi.org/project/pylint/#history) today (v2.14.0): perhaps they broke backward compatibility with that option. There are two experiments we can try:
- force pylint v2.13.9 in setup.cfg
- remove `# pylint: disable=no-self-use` from the code and see what happens

src/dali/transaction_resolver.py
stevendavis: Two blank lines between functions is standard.  My versions:
- black 22.3.0
- isort 5.9.3

These versions should also match the pre-commit configuration, however I see blacken-docs has a dependency on black 20.8b1.

eprbell: Actually you're right: the newline was missing.

eprbell (commented):
LGTM

eprbell:
Can you update your tree to pickup the fixes in #52? Then I think we can merge.

PR #54, opened 2022-07-09
Add ledger csv plugin.

stevendavis:
This is based on the trezor plugin.  Some notes:
1. The csv file contains transactions for multiple currencies.
2. I have only seen receive transactions.  The code supports send transactions, but it is untested.
3. There is a hard-coded dict to rename currencies.


src/dali/plugin/input/csv/ledger.py
```diff
@@ -0,0 +1,134 @@
...
+                if quantity_number == ZERO and fee_number > ZERO:
+                    self.__logger.warning("Possible dusting attack (fee > 0, total amount = 0): %s", raw_data)
+                    continue
+                if transaction_type in {_SENT, _RECV}:  # Need example data for sent transactions, untested as of 7/9/2022
+                    result.append(
+                        IntraTransaction(
+                            plugin=self.__LEDGER,
+                            unique_id=crypto_hash,
+                            raw_data=raw_data,
+                            timestamp=f"{timestamp_value}",
```
eprbell, line 119: Is this generated timestamp in [ISO8601 format](https://github.com/eprbell/dali-rp2/blob/main/docs/user_faq.md#what-is-the-timestamp-format)? Based on what you documented in line 16 I think so, but just wanted to double-check.

eprbell (approved):
Looks good, thanks for implementing this!

src/dali/plugin/input/csv/ledger.py
stevendavis: The timestamp in the Ledger CSV file appears to be ISO8601 format.  The date and time are always separated by "T", there is always a 3 digit millisecond fraction, and always a "Z" time zone designator, for example:
2022-06-05T00:39:14.007Z

The timezone string is parsed by the dateutil module on line 99, producing a datetime. Here we explicitly expect ISO8601 format.

The datetime is converted back to a string on line 119, but this is not ISO8601 format.  It will look like this:
2022-06-05 00:39:14.007000+00:00

Then, the AbstractTransaction uses the dateutil module to parse the string again without assuming any particular format.

PR #55, opened 2022-07-11
Breadth First Search Tool

macanudo527:
Breadth First Search Tool with trace. It currently only supports cyclic search and not tree.


macanudo527:
It took 4 commits, but I did narrow down the files to just what is needed.

eprbell (changes requested):
I was thinking of a generic Graph utility with these features:
* an encapsulated Graph class (not exposing the internal data structures it uses, but implemented with adjacency lists): this class would have an add_edge method to populate the graph
* an encapsulated Vertex class with Generics-based user data
* BFS, Dijkstra, etc. would be methods of Graph

However I don't want us to get too sidetracked to implement this. I propose the following:
* for the time being let's not do the generic graph utility 
* let's bring this BFS code inside the ccxt pair converter file (no need for a separate file for the time being)

Sorry about the back and forth. We can revisit a full graph API later.

macanudo527:
All right I'll port the code over in the next commit. Closing this PR.

PR #58, opened 2022-07-19
Add JP Plugin

macanudo527:
The JP plugin for dali-rp2


eprbell (approved):
LGTM!

eprbell:
The static analysis error is due to the fact that the JP country plugin on the RP2 side is not yet in an official release. I will make a new RP2 release tonight and the error should disappear.

eprbell:
I released new versions of both RP2 and DaLI, so the problem is not fixed and tests are passing: merging the code.

eprbell:
Yay, first non-US country officially supported!

PR #53, opened 2022-06-14
Ccxt pair converter

macanudo527:
This is the first draft of the CCXT pair converter. I hope all the features make sense. I tried my best to have the logic make sense, but the 'bridge pricing' (using bridge fiat and bridge crypto assets) might be a little confusing. Things might need to be refactored a little. Let me know if you have questions.

PR #60, opened 2022-07-23
Nexo plugin

iloveitaly:
Very rough right now: wanted to submit early to get some feedback on the approach.


macanudo527:
You might want to check the CCXT converter plugin #53 . It pulls historical data from Binance.com as well as any CCXT supported exchange. Right now, it pulls historical data specifically from the exchange listed in the historical bar, but we could easily edit it to default to Binance or set a default for Nexo since it looks like they don't have a pricing API. I'm sure @eprbell could weigh in on the best option. 

Either way I'm pretty sure there isn't a need for a separate converter plugin. It's currently undergoing final testing. I'd appreciate any input you have.

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,150 @@
...
+
+                            transaction_type=Keyword.INTEREST.name,
+                            # nexo does give us the spot price, but it's often 0 if subcent
+                            spot_price=Keyword.UNKNOWN.value,
+                            crypto_in=amount,
+                            fiat_fee="0",
+                        )
+                    )
+                elif transaction_type in [_LOCKING_TERM_DEPOSIT, _UNLOCKING_TERM_DEPOSIT]:
+                    # I don't think we need to record locking/unlocking deposits for term interest
```
eprbell, line 121: What are locking / unlocking deposits?

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,150 @@
...
+
+            for line in lines:
+                raw_data: str = self.__DELIMITER.join(line)
+                self.__logger.debug("Transaction: %s", raw_data)
+
+                transaction_id: str = line[self.__TRANSACTION_ID_INDEX].strip()
+                transaction_type: str = line[self.__TRANSACTION_TYPE_INDEX].strip()
+                currency: str = line[self.__CURRENCY_INDEX].strip()
+                amount = line[self.__AMOUNT_INDEX].strip()
+                timestamp_with_timezone = f"{line[self.__TIMESTAMP_INDEX].strip()} -00:00"
```
eprbell, line 100: This line suggests the timestamp in the CSV has no timezone information, but the timezone is always assumed to be UTC. Is this correct?

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,150 @@
...
+                            plugin=self.__NEXO,
+                            unique_id=transaction_id,
+                            raw_data=raw_data,
+                            exchange=self.__NEXO,
+                            holder=self.account_holder,
+                            timestamp=timestamp_with_timezone,
+                            asset=currency,
+
+                            transaction_type=Keyword.INTEREST.name,
+                            # nexo does give us the spot price, but it's often 0 if subcent
```
eprbell, line 114: If the price is given and it's > 0 we should use it. If the price is bad (e.g. 0 or missing) we can use `Keyword.UNKNOWN`.

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,150 @@
...
+
+                            to_exchange=self.__NEXO,
+                            to_holder=self.account_holder,
+
+                            # we do know the spot price, but nexo seems to do some aggressive rounding
+                            spot_price=Keyword.UNKNOWN.value,
+                            crypto_sent=Keyword.UNKNOWN.value,
+                        )
+                    )
+                else:
```
eprbell, line 146: Withdrawal transactions seem to be missing: if you don't have them in your own data, it's OK not to implement them (hopefully we'll get a user to open an issue as per the error log at line 147). Maybe just add a comment.

src/dali/plugin/pair_converter/binance.py
```diff
@@ -0,0 +1,78 @@
...
+                "2434.19055334",    // Quote asset volume
+                308,                // Number of trades
+                "1756.87402397",    // Taker buy base asset volume
+                "28.46694368",      // Taker buy quote asset volume
+                "17928899.62484339" // Ignore.
+            ]
+        ]
+        """
+
+        # binance has more tokens in USDT, which should be equal to the USD price
```
eprbell, line 54: I don't think this is a safe assumption: USDT oscillates around the peg, most of the time slightly, sometimes quite a bit (https://www.coinbase.com/price/tether).

src/dali/plugin/pair_converter/binance.py
```diff
@@ -0,0 +1,78 @@
...
+        if to_asset == "USD":
+            to_asset = "USDT"
+
+        symbol = f"{from_asset}{to_asset}"
+        twelve_hours = 60 * 60 * 12
```
eprbell, line 59: Could you make it into a global constant starting with underscore: `_TWELVE_HOURS`

src/dali/plugin/pair_converter/binance.py
```diff
@@ -0,0 +1,78 @@
...
+        utc_timestamp = int(timestamp.astimezone(timezone.utc).timestamp())
+
+        response = requests.get(f"https://api.binance.com/api/v3/klines?symbol={symbol}&interval=12h&startTime={utc_timestamp}&limit=1")
+        response_json = json.loads(response.text)
```
eprbell, line 63: Do we need any error checks here (e.g. [see this example](https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/rest/coinbase.py#L830))?

src/dali/plugin/pair_converter/binance.py
```diff
@@ -0,0 +1,78 @@
...
+
+        return HistoricalBar(
```
eprbell, line 65: Instead of hardcoded values (0, 1, 2, 3, etc.) please use global constants (uppercase and starting with underscore). E.g.: https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/rest/coinbase.py#L44

src/dali/in_transaction.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         asset: str,
         exchange: str,
         holder: str,
+        # TODO not sure why we aren't using `Keyword` here?
```
eprbell, line 39: Could you elaborate? Not sure I understand the comment.

src/dali/dali_main.py
```diff
@@ -163,6 +164,7 @@ def _dali_main_internal(country: AbstractCountry) -> None:
 
         if not pair_converter_list:
             pair_converter_list.append(HistoricCryptoPairConverterPlugin(Keyword.HISTORICAL_PRICE_HIGH.value))
+            pair_converter_list.append(BinancePairConverterPlugin(Keyword.HISTORICAL_PRICE_HIGH.value))
             LOGGER.info("No pair converter plugins found in configuration file: using Historic_Crypto/high as default")
```
eprbell, line 168: Change the message to "using default configuration".

eprbell (changes requested):
Thanks for submitting new plugins! The code looks pretty good and I just have a few nits and comments. As @macanudo527 mentioned, the Binance pair converter is probably overlapping with the CCXT pair converter he's working on (so we may not need it), but the Nexo CSV input plugin can definitely be merged.

Could you also add a small section for Nexo in the [docs](https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md#data-loader-plugin-sections)?

src/dali/dali_main.py
macanudo527: Sorry to butt in here, but should I add the CCXT converter to the default config like this as well?

eprbell: No worries! Yes, I think the CCXT converter should definitely be added to the default configuration. Currently we only have one converter (HistoricCrypto), so the default configuration is quite meager. We do need to make the default configuration better and adding CCXT is a step in the right direction.

eprbell:
> 

Actually this is an interesting idea: would it make sense to add variants (subclasses) of your CCXT converter that use a fixed exchange (one for Binance, one for Kraken, etc.)? The binance subclass would be exactly equivalent to the pair converter in this PR. Additionally, if we added a CCXT converter subclass for Coinbase, it should be equivalent to the HistoricCrypto plugin. Does that sound correct or am I missing something?

macanudo527:
The CCXT converter plugin reads in the exchange name from each transaction that is fed to it then pulls data from that exchange per transaction. 

So in theory, if you transacted on Binance, kraken and Coinbase it would pull pricing data from each exchange all in one plugin. 

I don't know what the use case would be for a converter 'locked' to one exchange. As we discussed before, there are price fluctuations between exchanges so it is best to pull pricing data from the specific exchange the crypto asset was transacted on. 

If done right, the CCXT converter should cover the vast majority of cases except for fringe exchanges with apis not supported by CCXT.

macanudo527:
Thinking on this more, what we could do is create a USD base pair version that only works on exchanges that have USD base pairs and for users that need USD denominated prices. 

We could rip out the routing / graphing and speed things up. 

Then, have a more robust version that handles everything.

eprbell:
> 

I understand the behavior of the CCXT plugin, reading the exchange from each transaction and that is great as it is. The "locked" behavior would offer an alternative/additional model. Each of the locked plugins would behave like the existing HistoricCrypto plugin (which ignores the exchange hint). The user could then arrange the locked plugins in a static configuration, defining a custom priority chain (e.g. try binance first, then coinbase, then kraken, etc). I am not sure if some users would prefer this, but it's a different model, with different operational semantics.

eprbell:
> 

Let me paint a scenario and ask a question:
* Bob is a US user (using the US country plugin) 
* he uses exchanges that have USD-based pairs
* he uses the CCXT plugin (as it is written today in your PR)

In the above situation, does the routing code get executed or not? I was under the impression that it would not be executed, but correct me if I'm wrong. If I'm right, would we still need a router-less variant of the CCXT exchange?

macanudo527:
No it doesn't. If the market exists for the conversion on the exchange:

Conversion BTC -> USD
Market BTC/USD 

Then, the routing isn't used. As is, the graph gets loaded but I'm going to push that down so it lazy loads. I guess it wouldn't offer too much of an efficiency boost.

eprbell:
> No it doesn't. If the market exists for the conversion on the exchange:
> 
> Conversion BTC -> USD Market BTC/USD
> 
> Then, the routing isn't used. As is, the graph gets loaded but I'm going to push that down so it lazy loads. I guess it wouldn't offer too much of an efficiency boost.

Ok, that's what I thought: sounds good.

macanudo527:
> > 
> 
> I understand the behavior of the CCXT plugin, reading the exchange from each transaction and that is great as it is. The "locked" behavior would offer an alternative/additional model. Each of the locked plugins would behave like the existing HistoricCrypto plugin (which ignores the exchange hint). The user could then arrange the locked plugins in a static configuration, defining a custom priority chain (e.g. try binance first, then coinbase, then kraken, etc). I am not sure if some users would prefer this, but it's a different model, with different operational semantics.

Ok, I got you now. I was misunderstanding this in thinking you wanted it for efficiency reasons. For priority reasons would probably be useful for some users that have assets on hardware wallets or some other place that doesn't have readily available pricing data. I opened an issue #61 so that this doesn't get lost.

src/dali/in_transaction.py
iloveitaly: The `configuration.Keyword` enum looks like what the input values should be here, curious what this isn't used as the type instead of a raw string.

src/dali/plugin/input/csv/nexo.py
iloveitaly: Got it, I'll add a comment. I don't have them in my own data.

src/dali/plugin/input/csv/nexo.py
iloveitaly: The price only has two decimal precision. Maybe I should check for 0 and use `UNKNOWN` in that case, otherwise use the provided price even with only 2 precision? Wdyt?

src/dali/plugin/input/csv/nexo.py
iloveitaly: I think that's right. I'll confirm and document in-code.

src/dali/plugin/input/csv/nexo.py
iloveitaly: I'll write an inline comment on this, but this is a Nexo thing: they "lock" your crypto in a "fixed term" deposit which earns higher interest.

i.e. these transactions just indicate that you cannot withdraw these funds while these are locked. So they effect your available balance.

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,150 @@
...
+                            # most likely, it's you, but we can't say for sure
+                            from_exchange=Keyword.UNKNOWN.value,
+                            from_holder=Keyword.UNKNOWN.value,
+
+                            to_exchange=self.__NEXO,
+                            to_holder=self.account_holder,
+
+                            # we do know the spot price, but nexo seems to do some aggressive rounding
+                            spot_price=Keyword.UNKNOWN.value,
+                            crypto_sent=Keyword.UNKNOWN.value,
```
iloveitaly, line 143: @eprbell this line actually caused me issues with `rp2`. Should I use `0.0` here instead? That seemed to solve the issue when I was trying to get rp2 to generate.

src/dali/plugin/input/csv/nexo.py
eprbell: That sounds correct. The Coinbase plugin had the same problem (two decimal precision): see https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/rest/coinbase.py#L723

src/dali/plugin/input/csv/nexo.py
eprbell: What was the issue: could you paste a stack trace?

src/dali/in_transaction.py
eprbell: Most of the parameters to instantiate a transaction are strings (they come from REST/CSV sources or from the configuration file), so the design choice has been to pass in strings and then [validate them inside the transaction constructor](https://github.com/eprbell/dali-rp2/blob/main/src/dali/in_transaction.py#L61). 

However it's true that for `transaction_type` we could have used an enum, because `transaction_type` is assigned in the code and not read from external input. I guess I chose string for uniformity with the rest of the parameters.

Note that the `transaction_type` parameter is [validated fully at runtime](https://github.com/eprbell/dali-rp2/blob/main/src/dali/in_transaction.py#L63) and even if we changed transaction_type to enum, we would still need to [validate the value](https://github.com/eprbell/dali-rp2/blob/5f91dd3c8fc9d4548ea9f9112936b97978e4decd/src/dali/in_transaction.py#L26), depending on the actual class type of the transaction (see the [DIRECTION_2_TRANSACTION_TYPE_SET](https://github.com/eprbell/dali-rp2/blob/main/src/dali/configuration.py#L125) dictionary) so not much would change in the code (we would only skip the string check).

eprbell:
> > > 
> > 
> > 
> > I understand the behavior of the CCXT plugin, reading the exchange from each transaction and that is great as it is. The "locked" behavior would offer an alternative/additional model. Each of the locked plugins would behave like the existing HistoricCrypto plugin (which ignores the exchange hint). The user could then arrange the locked plugins in a static configuration, defining a custom priority chain (e.g. try binance first, then coinbase, then kraken, etc). I am not sure if some users would prefer this, but it's a different model, with different operational semantics.
> 
> Ok, I got you now. I was misunderstanding this in thinking you wanted it for efficiency reasons. For priority reasons would probably be useful for some users that have assets on hardware wallets or some other place that doesn't have readily available pricing data. I opened an issue #61 so that this doesn't get lost.

Yes, that's a good example. E.g. if the user moves coins from Trezor to Ledger the "locked" instances of CCXT would be useful.

src/dali/plugin/input/csv/nexo.py
iloveitaly: Here's the error message on the rp2 side:

```
Exception: Encountered an unresolved DaLI transaction (read DaLI's documentation / FAQ to learn how to resolve this issue): 
```

When I filled in the `crypto_sent` value with `0.0` the issue was fixed. Any ideas here? In this case we are just receiving crypto, so it makes sense that the sent would be zero?

iloveitaly:
I've handled most of the comments on the nexo plugin. Using the CCTX-based pair makes sense instead of my hacky converter. Will take a look at that next!

eprbell (changes requested):
Looking better and better! As for the locked CCXT converter variants, I think they will be trivial subclasses of the CCXT converter: @macanudo527 should be able to tackle them when his CCXT converter is finished. So for now I think we can just remove the binance converter from this PR. But if you'd like to contribute more code there is plenty to do. Feel free to check open issues in DaLI and RP2, look at anything that interests you in the code, or just ask here for suggestions.

Also remember to add a small section for Nexo in the [docs](https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md#data-loader-plugin-sections): you can use the Ledger section as boilerplate.

src/dali/plugin/input/csv/nexo.py
```diff
@@ -97,53 +82,63 @@ def load(self) -> List[AbstractTransaction]:
...
+
+                            'transaction_type':Keyword.INTEREST.name,
+                            'spot_price': spot_price,
+                            'crypto_in': amount,
+                            'fiat_fee': "0",
+                        }))
                     )
                 elif transaction_type in [_LOCKING_TERM_DEPOSIT, _UNLOCKING_TERM_DEPOSIT]:
+                    # These are unique to Nexo thing: they "lock" your crypto in a "fixed term" deposit which earns higher interest.
+                    # i.e. these transactions just indicate that you cannot withdraw these funds while these are locked. So they effect your available balance.
```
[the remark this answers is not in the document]
iloveitaly: That's correct—they are just transfers within Nexo. They cannot lock funds from other wallets/accounts.

src/dali/plugin/input/csv/nexo.py
```diff
@@ -97,53 +82,63 @@ def load(self) -> List[AbstractTransaction]:
...
-                            unique_id=transaction_id,
-                            raw_data=raw_data,
-                            timestamp=timestamp_with_timezone,
-                            asset=currency,
-
-                            crypto_received=amount,
+                        IntraTransaction(**(common_params | {
+                            'crypto_received':amount,
 
                             # most likely, it's you, but we can't say for sure
```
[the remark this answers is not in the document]
iloveitaly: Cleared this up. I think we should keep the comment—for new devs coming to the codebase it makes it easier to understand _why_ we are using unknown, otherwise it seems like an error/omission.

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,145 @@
...
+                            'crypto_received':amount,
+
+                            # most likely, it's you, but we can't say for sure
+                            'from_exchange':Keyword.UNKNOWN.value,
+                            'from_holder':Keyword.UNKNOWN.value,
+
+                            'to_exchange':self.__NEXO,
+                            'to_holder':self.account_holder,
+
+                            # we do know the spot price, but nexo seems to do some aggressive rounding
```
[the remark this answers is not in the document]
iloveitaly: Good call!

src/dali/transaction_resolver.py
```diff
@@ -142,7 +142,7 @@ def _update_spot_price_from_web(transaction: AbstractTransaction, global_configu
         )
         if conversion.rate is None:
             raise Exception(
-                f"Spot price for {transaction.timestamp_value}:{transaction.asset}->{global_configuration[Keyword.NATIVE_FIAT.value]}"
+                f"Spot price for {transaction.unique_id}:{transaction.timestamp_value}:{transaction.asset}->{global_configuration[Keyword.NATIVE_FIAT.value]}"
```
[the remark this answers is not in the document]
iloveitaly: Are you cool with just removing the colon altogether? Is there an easy way to do this without messy nil-check logic?

iloveitaly:
@eprbell I removed the binance pair converter with the idea that we'll get the other PR merged and can leverage that. Addressed most of the PR comments. Thanks for the thorough review!

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,151 @@
...
+                    self.__logger.debug("Skipping lock or unlock deposit: %s", line)
+                elif transaction_type == _DEPOSIT:
+                    result.append(
+                        IntraTransaction(
+                            **(
+                                common_params
+                                | {
+                                    "crypto_received": amount,
+                                    # most likely, funds are coming from the user/tax payer, but we can't say for sure so we use unknown
+                                    # and let the user manually input the owner of these funds.
```
eprbell, line 135: Actually the missing information is filled automatically by the transaction resolver:
```suggestion
                                    # and let the DaLI transaction resolver fill in the missing details.
```

src/dali/transaction_resolver.py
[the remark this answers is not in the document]
eprbell: If we just removed the colon unique_id and timestamp_value would be printed together, which would make it hard to read. How about:
```suggestion
                f"Spot price for {transaction.unique_id + ':' if transaction.unique_id else ''}:{transaction.timestamp_value}{transaction.asset}->{global_configuration[Keyword.NATIVE_FIAT.value]}"
```

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,151 @@
...
+                                    "transaction_type": Keyword.INTEREST.name,
+                                    "spot_price": spot_price,
+                                    "crypto_in": amount,
+                                    "fiat_fee": "0",
+                                }
+                            )
+                        )
+                    )
+                elif transaction_type in [_LOCKING_TERM_DEPOSIT, _UNLOCKING_TERM_DEPOSIT]:
+                    # These are unique to Nexo thing: they "lock" your crypto in a "fixed term" deposit which earns higher interest.
```
eprbell, line 123: Nit:
```suggestion
                    # These are unique to Nexo: they "lock" your crypto in a "fixed term" deposit which earns higher interest.
```

eprbell (changes requested):
Almost ready to merge. Here's what needs to be done:
* I added a few minor comments/nits to the latest version of the code;
* add a small section for Nexo in the [docs](https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md#data-loader-plugin-sections): you can use the Ledger section as boilerplate;
* fix static analysis errors.

iloveitaly:
I believe I fixed all of the linting errors! Will take care of doc updates later this week. Appreciate you working with me on this!

eprbell:
Thank you for submitting PRs! I will review the code for this and the other PRs in the next few days, however I wanted to share an important point that applies to all data loader plugins (so it applies #60, #65 and #66). It has to do with the `unique_id` field, which requires special attention. This field is used by the DaLI transaction resolver to match incomplete transaction and join them. E.g. a transfer from Coinbase to Trezor is initially represented as 2 transactions:
* an IntraTransaction from Coinbase to UNKNOWN, generated by the Coinbase plugin (which doesn't know that the destination is a Trezor)
* an IntraTransaction from UNKNOWN to Trezor, generated by the Trezor plugin (which doesn't know that the source is Coinbase)

These two transactions both share the same `unique_id` (which is typically the hash and is filled in by the respective plugins): they are fed to the transaction resolver, which uses `unique_id` to match them up and to join them into a single transaction without UNKNOWN fields. 

This is why `unique_id` is so important: if an incorrect value is used, the transaction resolver is unable to match and join incomplete transactions.

The reason I point this out is that in generated CSV files sometimes the transaction id is empty, other times (confusingly) it's an internal id of the exchange (which cannot be used by the transaction resolver). In such cases (where hash is not available) DaLI plugins must set the `unique_id` field to UNKNOWN.

Some examples of this:
* BlockFI CSVs don't contain the hash or any other transaction id
* Coinbase CSVs contain an internal transaction id, but it's not the hash

Check the following for more details:
*  [transaction resolver dev docs](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#the-transaction-resolver)
* [manual plugin docs](https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md#partial-transactions-and-transaction-resolution), which contain an example of how users would use the `unique_id` field in manual files to finish incomplete transactions for unsupported exchanges.

There is a way to tell if you're using the right `unique_id` in your plugins: e.g. if you have a transfer from Coinbase to Nexo or viceversa, make sure the final generated file contains only one transaction for this (with "from" fields set to Coinbase, "to" fields set to Nexo and no UNKNOWN source or destination).

This applies to all CSV and REST plugins, but it's especially relevant to CSV plugins (because REST APIs often provide hash information): make sure all these new plugins use either the transaction hash or UNKNOWN as the `unique_id`.

iloveitaly:
@eprbell Wow, thank you so much for the information! Your PR reviews are the most helpful I've ever gotten from an open source project, very impressed.

This makes sense: in other words, even if there is a unique ID in a CSV _we don't want to include it_ unless it's referenced in another data source (i.e. external id or foreign key). I'll adjust the PRs!

src/dali/plugin/input/csv/nexo.py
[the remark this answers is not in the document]
iloveitaly: Adding a comment to this effect.

iloveitaly:
Addressed comments + added documentation! Let me know if anything else is needed here.

eprbell:
> 

Hey, anytime! And thank you for submitting code and helping out with the project: two way street and all that :-D.

Correct, we don't want to include `unique_id` unless it's either the transaction hash (which is always understood by all exchanges and wallets) or in some special cases the internal id (e.g. transactions from Coinbase to Coinbase Pro don't use the transaction hash, but an internal Coinbase-only id which both CB and CB Pro understand).

docs/configuration_file.md
```diff
@@ -114,6 +114,21 @@ csv_file = <em>&lt;csv_file&gt;</em>
...
+```
+[dali.plugin.input.csv.nexo <qualifiers>]
+account_holder = <account_holder>
+account_nickname = <account_nickname>
+transaction_csv_file = <csv_file>
+```
+
+Notes:
+
+* There is a transaction ID in the Nexo file, but it's an internal transaction ID so it ignored
```
eprbell, line 129: I think this is not relevant to the user: it's more targeted to developers. I would remove it from the user docs and maybe add it back as a comment in the code.

src/dali/in_transaction.py
```diff
@@ -36,6 +36,9 @@ def __init__(
         asset: str,
         exchange: str,
         holder: str,
+        # validated to be a `configuration.Keyword` enum value at runtime
+        # string is used instead of a `Keyword` type to allow for more flexible input
+        # More info: https://github.com/eprbell/dali-rp2/pull/60#discussion_r928178970)
```
eprbell, line 41: I think this comment is not needed (and also not relevant to this PR): all parameters are validated at runime, not just this one (in out- and intra-transactions as well). Let's just remove it for now. If there is a need for additional clarification we can add a developer FAQ (in a separate PR).

src/dali/plugin/input/csv/nexo.py
```diff
@@ -0,0 +1,149 @@
...
+
+                # nexo does give us the spot price, but it's often 0 if a subcent value
+                # if it is non-zero, we use it, otherwise we use unknown as the value
+                # the spot price contains $ char, so we remove it
+                raw_spot_price = RP2Decimal(re.sub(r"[^\d.]", "", line[self.__SPOT_PRICE_INDEX]))
+
+                if raw_spot_price.is_zero():
+                    spot_price = Keyword.UNKNOWN.value
+                else:
+                    spot_price = str(raw_spot_price)
```
eprbell, line 102: Prefer ternary operator:
```suggestion
                spot_price = str(raw_spot_price) if not raw_spot_price.is_zero() else Keyword.UNKNOWN.value
```

src/dali/transaction_resolver.py
```diff
@@ -142,7 +142,8 @@ def _update_spot_price_from_web(transaction: AbstractTransaction, global_configu
         )
         if conversion.rate is None:
             raise Exception(
-                f"Spot price for {transaction.timestamp_value}:{transaction.asset}->{global_configuration[Keyword.NATIVE_FIAT.value]}"
+                f"Spot price for {transaction.unique_id + ':' if transaction.unique_id else ''}"
+                f"{transaction.timestamp_value}{transaction.asset}->{global_configuration[Keyword.NATIVE_FIAT.value]}"
```
eprbell, line 146: Missing `:` after `{transaction.timestamp_value}` (it's needed to separate fields)

eprbell (changes requested):
Here's my latest review.

src/dali/in_transaction.py
iloveitaly: Removed! For me as a new developer to the project, this was confusing (esp since typing is used pretty consistently through the codebase).

src/dali/transaction_resolver.py
iloveitaly: Done!

docs/configuration_file.md
iloveitaly: Done!

eprbell (approved):
Ship it!

PR #64, opened 2022-08-01
Support additional blockfi transactions

iloveitaly:
Adding support for additional blockfi transaction types in my csv export

### DIFF
--- a/src/dali/plugin/input/csv/blockfi.py
+++ b/src/dali/plugin/input/csv/blockfi.py
@@ -161,7 +161,7 @@ class InputPlugin(AbstractInputPlugin):
                             notes="ACH withdrawal",
                         )
                     )
-                elif transaction_type in [_WITHDRAWAL, _BIA_WITHDRAWAL]:
+                elif transaction_type == _WITHDRAWAL:
                     amount: RP2Decimal = RP2Decimal(line[self.__AMOUNT_INDEX])
                     amount = -amount  # type: ignore
                     if last_withdrawal_fee is not None:
@@ -204,6 +204,10 @@ class InputPlugin(AbstractInputPlugin):
                 elif transaction_type == _TRADE:
                     # Trades will be handled by parsing trade_report_all.csv
                     # export
+                    continue
+                elif transaction_type == _BIA_WITHDRAWAL:
+                    # these withdrawals are internal transfers within blockfi which is why they are skipped
+                    # https://github.com/eprbell/dali-rp2/pull/64
                     continue
                 elif transaction_type == _WITHDRAWAL_FEE:
                     last_withdrawal_fee = RP2Decimal(line[self.__AMOUNT_INDEX])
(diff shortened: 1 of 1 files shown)


src/dali/plugin/input/csv/blockfi.py
```diff
@@ -152,14 +154,14 @@ def load(self) -> List[AbstractTransaction]:
...
-                            transaction_type="Sell",
+                            transaction_type=Keyword.SELL.value,
                             spot_price=Keyword.UNKNOWN.value,
                             crypto_out_no_fee=str(-RP2Decimal(line[self.__AMOUNT_INDEX])),
                             crypto_fee="0",
                             notes="ACH withdrawal",
                         )
                     )
-                elif transaction_type == _WITHDRAWAL:
+                elif transaction_type in [_WITHDRAWAL, _BIA_WITHDRAWAL]:
```
eprbell, line 164: Just curious: what's the difference between BIA withdrawal and vanilla withdrawal?

eprbell (commented):
LGTM!

src/dali/plugin/input/csv/blockfi.py
eprbell: Oh, I just realized: it's withdrawal from their interest account.

eprbell: My understanding is that this type of transaction is just an internal transfer from the interest account to the normal account (at least that's how it shows in my case): if that's correct, this is not a real withdrawal, because the amount never leaves BlockFi (in this case it would be better to just ignore "BIA Withdraw"). Can you verify this?

eprbell (changes requested):
Let's verify the nature of BIA Withdraw (see my inline comment).

src/dali/plugin/input/csv/blockfi.py
iloveitaly: After looking at my CSV report more closely I'm seeing the same thing:

```
USDC,-21.19100000,Withdrawal,2022-06-15 20:29:37
USDC,21.19100000,BIA Withdraw,2022-06-14 00:52:19
USDC,21.19198526,Interest Payment,2022-05-31 23:59:59
```

I think we can safely just ignore and `DEBUG` log the BIA withdrawal since it is just a transfer between internal blockfi accounts, right?

iloveitaly (commented):
Updated PR to skip BIA withdrawals

src/dali/plugin/input/csv/blockfi.py
eprbell: I agree: we can ignore it for the purposes of RP2 calculations.

src/dali/plugin/input/csv/blockfi.py
```diff
@@ -203,6 +205,10 @@ def load(self) -> List[AbstractTransaction]:
                     # Trades will be handled by parsing trade_report_all.csv
                     # export
                     continue
+                elif transaction_type == _BIA_WITHDRAWAL:
+                    # these withdrawals are internal transfers within blockfi which is why they are skipped
+                    # https://github.com/eprbell/dali-rp2/pull/64
+                    continue
```
eprbell, line 211: Add a debug log with the raw data:
```
self.__logger.debug("BIA Withdraw: %s", raw_data)
```

eprbell (changes requested):
One minor nit and then it's ready to merge.

src/dali/plugin/input/csv/blockfi.py
iloveitaly: Done!

eprbell (approved):
LGTM!

PR #66, opened 2022-08-01
Coinbase CSV plugin

iloveitaly:


eprbell:
Thanks for submitting a PR! I will review this in the next few days, but meanwhile check my comments on `unique_id` in #60 (which apply here too).

eprbell (commented):
Looks to me this is parsing the "Reports / Visit Taxes / Raw Transactions Report" CSV file: is this correct? I looked at that file and it seems quite incomplete. In my generated file there are some Coinbase to Coinbase Pro transfers, some Coinbase Pro to Coinbase ones, but not all of them. Also external transfers seem to be missing altogether. As it is, I think may be too incomplete as a data source for RP2. 

A good test is to generate the RP2 report with the CSV plugin and then with the REST plugin and then compare the results (you can use [ods_diff.py](https://github.com/eprbell/rp2/blob/main/tests/ods_diff.py) utility to compare ODS files). 

Coinbase also generates another CSV file ("Reports / Transaction History"), which is also missing data: e.g. Coinbase Pro transfers and transaction hash. 

If the data source is too incomplete I think it may be best to just direct users to the REST plugin.

src/dali/plugin/input/csv/coinbase.py
```diff
@@ -0,0 +1,206 @@
...
+            self.__logger.debug("Header: %s", header)
+
+            for line in lines:
+                raw_data: str = self.__DELIMITER.join(line)
+                self.__logger.debug("Transaction: %s", raw_data)
+
+                transaction_id: str = line[self.__TRANSACTION_ID_INDEX].strip()
+                transaction_type: str = line[self.__TRANSACTION_TYPE_INDEX].strip()
+
+                # TODO unsure how to handle conversions, I think this is only for ETH2
```
eprbell, line 88: Check how the [REST Coinbase plugin](https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/rest/coinbase.py) does it (look for "swap" in the code)

src/dali/plugin/input/csv/coinbase.py
```diff
@@ -0,0 +1,206 @@
...
+                if transaction_type == _CONVERT:
+                    continue
+
+                # there is no timezone information in the CSV, so we assume UTC
```
eprbell, line 92: There is (at least in mine): the Z at the end of the timestamp means UTC

eprbell:
Since we already have a REST plugin and Coinbase CSV are fairly incomplete this is probably not needed.

PR #70, opened 2022-08-07
Documentation for Binance.com plugins, CCXT pair converter plugin, and JSON documentation

macanudo527:
I'd appreciate any input on the wording for the CCXT pair converter plugin especially. It was a bit difficult to describe succintly.


docs/configuration_file.md
```diff
@@ -17,13 +17,16 @@
 ## Table of Contents
 * **[Introduction](#introduction)**
 * **[Data Loader Plugin Sections](#data-loader-plugin-sections)**
+  * [Binance.com Section (REST)](#binance-com-section-rest)
   * [Coinbase Section (REST)](#coinbase-section-rest)
   * [Coinbase Pro Section (REST)](#coinbase-pro-section-rest)
+  * [Binance.com Section (CSV)](#binance.com-section-csv)
```
eprbell, line 23: Add supplemental

docs/configuration_file.md
```diff
@@ -65,6 +68,28 @@ Where:
 
 DaLI comes with a few builtin plugins, but more are needed: help us make DaLI a robust open-source, community-driven crypto data loader by [contributing](../CONTRIBUTING.md#contributing-to-the-repository) plugins for exchanges and wallets!
 
+### Binance.com Section (REST)
+This plugin is REST-based and requires setting up API Keys in your Binance.com account settings (click on the API link).
```
eprbell, line 72: This line is the same as the Coinbase one: just wanted to make sure this is not a copy/paste error and that Binance.com also has an "API" link in the settings.

docs/configuration_file.md
```diff
@@ -65,6 +68,28 @@ Where:
...
+account_holder = <em>&lt;account_holder&gt;</em>
+api_key = <em>&lt;api_key&gt;</em>
+api_secret = <em>&lt;api_secret&gt;</em>
+username = <em>&lt;username&gt;</em>
+native_fiat = <em>&lt;native_fiat&gt;</em>
+</pre>
+
+Notes: 
+* The `username` parameter is optional and denotes the username used when connecting to the Binance.com mining pool. Only basic mining deposits (type 2) are currently supported.
+* Due to a bug in the Binance.com system, there may be duplicate locked savings/staked and flexible savings dividends around May 8th, 2021. Please review your payments around this time before processing your transactions with RP2.
```
eprbell, line 90: Is there a bug tracker link on their site or somewhere else? It may be helpful to add it here.

docs/configuration_file.md
```diff
@@ -65,6 +68,28 @@ Where:
...
+* Due to the information not being available via REST, autoinvest trades and ETH to BETH conversions can not be processed with this plugin. Please download the CSV for these transactions and use the Binance.com CSV plugin.
```
eprbell, line 91: I remember you mentioned that these files need to be converted to CSV format: maybe mention that here if it's the case.

docs/configuration_file.md
```diff
@@ -102,6 +127,22 @@ thread_count = <em>&lt;thread_count&gt;</em>
 
 Note: the `thread_count` parameter is optional and denotes the number of parallel threads used to by the plugin to connect to the endpoint. The higher this number, the faster the execution, however if the number is too high the server may interrupt the connection with a rate-limit error.
 
+### Binance.com Section (CSV)
```
eprbell, line 130: Add "supplemental"

docs/configuration_file.md
eprbell: Add "supplemental"

docs/configuration_file.md
```diff
@@ -270,6 +311,37 @@ Where:
...
+Initialize this plugin section as follows:
+<pre>
+[dali.plugin.pair_converter.ccxt</em>]
+historical_price_type = <em>&lt;historical_price_type&gt;</em>
+fiat_priority = <em>&lt;fiat_priority&gt;</em>
+</pre>
+
+Where:
+* `<historical_price_type>` is one of `open`, `high`, `low`, `close`, `nearest`. When DaLi downloads historical market data, it captures a `bar` of data surrounding the timestamp of the transaction. Each bar has a starting timestamp, an ending timestamp, and OHLC prices. You can choose which price to select for price lookups. The open, high, low, and close prices are self-explanatory. The `nearest` price is either the open price or the close price of the bar depending on whether the transaction time is nearer the bar starting time or the bar ending time.
+* `fiat_priority` is an optional list of strings in JSON format that ranks the priority of fiat in the routing system. If no `fiat_priority` is given, the default priority is `["USD", "JPY", "KRW", "EUR", "GBP", "AUD"]`, which is based on the volume of the fiat market paired with BTC (ie. BTC/USD has the highest worldwide volume, than BTC/JPY, etc.).
```
eprbell, line 326: Can you mention the string format: is it comma-separated, space-separated, etc. Also update the default priority from JSON-list to the same format seen by the user.

eprbell: If we mention the default list here, we need to remember to update it, if we update it in the code: add a comment in the code reminding to update docs.

eprbell: than -> then

docs/configuration_file.md
```diff
@@ -270,6 +311,37 @@ Where:
...
+
+The CCXT pair converter plugin uses a routing system to find the shortest pricing path between a base asset and a quote asset (what the asset is priced in). It does this by assembling a graph of nodes made out of assets and edges made from markets with a preference for the exchange the asset was purchased on. Fiat exchange rates from the European Central Bank are also added to the graph to allow any fiat to be converted between each other. 
+
+For example, if a user needs the price of their BETH (Beaconed ETH) purchase on Binance.com in CHF (Swiss Francs). The router will first route the price through the available markets on the exchange:
```
eprbell, line 330: purchased

eprbell (changes requested):
Awesome stuff!

docs/configuration_file.md
```diff
@@ -270,6 +311,37 @@ Where:
...
+
+1. BETH -> ETH
+2. ETH -> GBP
+
+Then, it will route the price through a fiat conversion to get the final price:
+
+3. GBP -> CHF
```
eprbell, line 337: The example is very helpful. Maybe describe what fiat_priority is being used (the default one? ) and add a brief comment in each line describing what is happening:
e.g. in the case of ETH -> GBP, why is it picking GBP (which seems quite low in the default list)?

docs/configuration_file.md
macanudo527: I clarified this a bit, thanks.

docs/configuration_file.md
macanudo527: I gave an example of JSON format which hopefully makes it clear.

docs/configuration_file.md
macanudo527: Are you talking about line 330? I was using the noun, but maybe the adjective is clearer.

docs/configuration_file.md
macanudo527: This is just a bug I've discovered on my own. Let me try to find if it is documented anywhere.

docs/configuration_file.md
macanudo527: I also added a comment to the CCXT pair converter PR #53

docs/configuration_file.md
eprbell: Sounds good, but don't spend a lot of time searching for it. If you don't have an official link to the bug, we can move on.

docs/configuration_file.md
eprbell: The string in JSON format is what was confusing me, because it's in the context of a .ini file. I was thinking we could use a comma-separated list of values without white space, but it's not terribly pretty: let's go with your JSON line.

eprbell (commented):
I think this is also good to merge. So I think the only that is missing is #53: let me know when that is ready and I'll review it from scratch one last time. Then it's merge time!

docs/configuration_file.md
macanudo527: I asked them and they said the discrepancy is based on the old system timestamp is when the interest is technically delivered, but the new system reports when the interest affects the user's wallet. I added the link in the docs.

eprbell (approved):
I don't have any more feedback on this: ready to merge.

PR #73, opened 2022-08-13
Nexo: corrected spot price, unique id, and safer float extraction

iloveitaly:
* Spot price calculation was incorrect. The price given by nexo is the realized usd amount, not the spot price per unit of asset
* The transaction hash is included in the description for transfers!
* Sometimes nexo uses scientific notation, so only $ are stripped instead of non-number values

### DIFF
--- a/src/dali/plugin/input/csv/nexo.py
+++ b/src/dali/plugin/input/csv/nexo.py
@@ -121,10 +121,12 @@ class InputPlugin(AbstractInputPlugin):
                     # I don't think we need to record locking/unlocking deposits for term interest
                     self.__logger.debug("Skipping lock or unlock deposit: %s", line)
                 elif transaction_type == _DEPOSIT:
+                    unique_id = common_params["unique_id"]
+
                     # nexo includes a network transaction hash in the case of deposits: it's embedded in the details field
                     # let's extract it using a regex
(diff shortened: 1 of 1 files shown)


src/dali/plugin/input/csv/nexo.py
```diff
@@ -120,11 +121,19 @@ def load(self) -> List[AbstractTransaction]:
                     # I don't think we need to record locking/unlocking deposits for term interest
                     self.__logger.debug("Skipping lock or unlock deposit: %s", line)
                 elif transaction_type == _DEPOSIT:
+                    unique_id = common_params["unique_id"]
+
+                    # nexo includes a network transaction hash in the case of deposits: it's embedded in the details field
```
eprbell, line 126: Crazy huh? I'm a little concerned about extracting data from an unstructured comment (because it can change at anytime). I guess let's leave that in for now and see how it goes: hopefully we can upgrade to a REST-based plugin for Nexo at some point and we won't have to worry about it.

eprbell (approved):
LGTM

src/dali/plugin/input/csv/nexo.py
iloveitaly: Yeah, super crazy. I hate it too but it's better than not having the uid, and I'm hoping relying on the char count is semi-reliable.

PR #74, opened 2022-08-13
Throw an exception when a manual csv does not have enough columns

iloveitaly:
I interpretted the "optional" columns to be optional within the CSV structure. This change
provides clear feedback to users if they do not include enough columns.

### DIFF
--- a/src/dali/plugin/input/csv/manual.py
+++ b/src/dali/plugin/input/csv/manual.py
@@ -114,7 +114,7 @@ class InputPlugin(AbstractInputPlugin):
             for line in lines:
                 raw_data: str = ",".join(line).strip()
                 if not header_found:
-                    # let's let the user know they don't have enough columns
+                    # let user know there is not enough columns
                     if len(line) - 1 < self.__IN_NOTES_INDEX:
                         raise ValueError(f"Not enough columns: the CSV must contain {self.__IN_NOTES_INDEX} columns.")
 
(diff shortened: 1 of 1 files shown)


src/dali/plugin/input/csv/manual.py
```diff
@@ -114,6 +114,10 @@ def _load_in_file(self, transactions: List[AbstractTransaction]) -> None:
             for line in lines:
                 raw_data: str = ",".join(line).strip()
                 if not header_found:
+                    # let's let the user know they don't have enough columns
+                    if len(line) - 1 < self.__IN_NOTES_INDEX:
+                        raise ValueError("Not enough columns, the CSV must contain all columns, even optional columns.")
```
eprbell, line 119: Nice! Can you add the same logic to out and intra?

eprbell: Also let's tell users how many columns are needed:
```suggestion
                        raise ValueError(f"Not enough columns: the CSV must contain {self.__IN_NOTES_INDEX} columns.")
```

eprbell (changes requested):

iloveitaly:
[Does this line](https://github.com/iloveitaly/dali-rp2/blob/da643f9dc08a19451aaae20dcf742fdea6049683/src/dali/transaction_resolver.py#L634) trigger your memory?

Maybe explicitly defining a fee as zero ensures that the sent amount is accurate. Ensures a level of correctness about the out flow which in doesn't need since the net amount settled to an account is net of fees?

iloveitaly:
@eprbell done! :)

src/dali/plugin/input/csv/manual.py
```diff
@@ -114,6 +114,10 @@ def _load_in_file(self, transactions: List[AbstractTransaction]) -> None:
             for line in lines:
                 raw_data: str = ",".join(line).strip()
                 if not header_found:
+                    # let user know there is not enough columns
+                    if len(line) - 1 < self.__IN_NOTES_INDEX:
+                        raise ValueError(f"Not enough columns: the CSV must contain {self.__IN_NOTES_INDEX} columns.")
```
eprbell, line 119:
```suggestion
                        raise ValueError(f"Not enough columns: the {self.__in_csv_file} CSV must contain {self.__IN_NOTES_INDEX} columns.")
```

eprbell (changes requested):

eprbell (approved):
LGTM!

PR #75, opened 2022-08-13
Don't prevent empty on out transactions

iloveitaly:
They are not preventing on in transactions. I didn't see any reason why we shouldn't use the same field requirements.

### DIFF
--- a/src/dali/out_transaction.py
+++ b/src/dali/out_transaction.py
@@ -39,7 +39,7 @@ class OutTransaction(AbstractTransaction):
         transaction_type: str,
         spot_price: str,
         crypto_out_no_fee: str,
-        crypto_fee: str,
+        crypto_fee: Optional[str] = None,
         crypto_out_with_fee: Optional[str] = None,
         fiat_out_no_fee: Optional[str] = None,
         fiat_fee: Optional[str] = None,
@@ -123,7 +123,7 @@ class OutTransaction(AbstractTransaction):
         return self.__crypto_out_no_fee
 
     @property
-    def crypto_fee(self) -> str:
+    def crypto_fee(self) -> Optional[str]:
         return self.__crypto_fee
 
     @property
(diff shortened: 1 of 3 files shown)


eprbell (changes requested):
In the early iterations of RP2 and DaLI fees were the subject of intense discussion and brainstormings. The resulting transaction constructor signatures are deliberate and the result of those discussions, however, I can't remember exactly why fees are optional in in-transactions but not in out-transactions. If you'd like to explore this here are a few things to keep in mind:
* update the constructor signature and the crypto fee property to have optional type (this is what the static anaysis is pointing out). This will have various ramifications in the code, I believe (static analysis will point out many of them);
* RP2 transaction constructor signatures must be consistent with DaLI's so you'll need to update the OutTransaction constructor and the crypto_fee property there too, which will have more ramifications.

As long as it doesn't break any tests we can consider doing this change, but let's revisit this once we understand how much code has to change in both projects.

iloveitaly:
[Does this line](https://github.com/iloveitaly/dali-rp2/blob/da643f9dc08a19451aaae20dcf742fdea6049683/src/dali/transaction_resolver.py#L634) trigger your memory?

Maybe explicitly defining a fee as zero ensures that the sent amount is accurate. Ensures a level of correctness about the out flow which in doesn't need since the net amount settled to an account is net of fees?

eprbell:
I thought some more about this. While I still can remember exactly the point in the discussion where this decision was made, I think the reasoning was more philosophical than anything else: 
* in in-transactions the fee sometimes makes sense as a concept and sometimes it doesn't. E.g. BUY needs the fee (which could be 0 but still makes sense in that context), but INTEREST or INCOME don't need the fee altogether (it doesn't make sense as a concept in that context). To capture this the fee is optional;
* in out-transactions the fee is always needed (again, it could be zero but it's always needed as a concept). There are no OUT transactions for which the fee is not needed as a concept. To capture this the fee is non-optional. 

It's not a super-strong point, but this was probably the thinking.

PR #77, opened 2022-08-16
Manual price converter

iloveitaly:
I'm just about out of time to contribute to this project, but here's the manual price converter that I put together. Obviously it's really rough and needs some work,
but hopefully it helps provide a starting point.

I also think it makes sense to include a logger on the abstract class to each subclass doesn't need to construct it's own.


iloveitaly:
@eprbell what did you think of this one?

eprbell:
> @eprbell what did you think of this one?

I'm OK with exploring this as we discussed in #53: using a .csv is probably the best approach.

Issue #22, opened 2022-04-10
 Implement CCXT-based Data Loader Abstract Plugin


eprbell:
Writing a DaLI data loader plugin has high impact on the functionality and usefulness of both [DaLI](https://github.com/eprbell/dali-rp2) and [RP2](https://github.com/eprbell/rp2). Data loader plugins are well-defined, encapsulated modules that translate native exchange REST API-based data into DaLI's standard format. As such they are good first issues for newcomers to the the project.

Before starting, please read the [contributing guidelines](https://github.com/eprbell/dali-rp2/blob/main/CONTRIBUTING.md#contributing-to-the-repository).

Plugin development is described in the [developer documentation](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md), in particular read the [Dali Internals](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#dali-internals) and [Plugin Development](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#plugin-development) sections, which contain all the information needed to build a new data loader plugin.

CCXT is documented here: https://docs.ccxt.com/en/latest/manual.html. This would be an abstract plugin that is meant to be subclassed by concrete ones.

The [Coinbase plugin](https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/rest/coinbase.py) can be used as an example.

Before submitting a PR for the new plugin, make sure to go through the [DaLI Plugin Laundry List](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#plugin-laundry-list).

jamesbaber1:
Ill can take a stab at this. When you say abstract, your still leaving this plugin the ability to be subclassed right?  Cause while ccxt does a good job generalizing a lot of exchanges, not everything can 100% be guaranteed.  For example trade history papigation might need to be changed on kucoin for example, which is just a few extra attributes on the ccxt config, or the REST api versions need to switch to get older data etc. 

So how I imagine this plugin would work is either as abstract class or the actual class.  For example wed right a exchange specific subclass of it that we know 100% will work, and we make a list saying these are the supported exchanges.  Then we say that if your exchange doesn't have a exchange specific implementation of ccxt, the fall back is this, however there is no guarantee on our end that RP2 will work on this exchange, however it is quite possible it will.

eprbell:
[...]

PR #80, opened 2022-09-04
Ccxt Input Abstract Plugin

macanudo527:
This PR addresses issue #22 .

This only has `_process_trades()` functioning. I commented out code that I'll edit later. 

Basically, I just want to submit this to see if this is the architecture you were thinking and if everything looks right. If it looks good to you, I'll continue on with deposits, withdrawals, and then the implicit API.

I think I can implement pagebased pagination from the Coinbase code, but idbased might be difficult to do without a real-world example.

Let me know what you think.


eprbell (changes requested):
This is cool. As I was reading the code I got a new design insight: I realized that pagination loop logic should probably be captured using custom iterators. It's a more natural fit and it's probably worth spending a little time exploring this approach. Custom iterators are already used extensively in the rest of the code. For a simple example of custom iterator see https://github.com/eprbell/rp2/blob/main/src/rp2/abstract_accounting_method.py.

The idea is that there is a container class (`AcquiredLotCandidates`) we want to iterate on, so we define an `__iter__` method which returns a custom iterator on that container (`AccountingMethodIterator`). The iterator class is custom defined: what `AccountingMethodIterator` a Python iterator is the `__next__` method, which returns the next element (in this case it's an InTransaction). When there are no more elements to iterate on `__next__` raises `StopIteration`.

So translating the above example to our case here:
* `AbstractPaginationDetails` becomes `AbstractPaginationDetailSet`, `DateBasedPaginationDetails` becomes `DateBasedPaginationDetailSet`, etc. (to capture the idea that they are container classes)
* Add "element" classes `AbstractPaginationDetails`, `DateBasedPaginationDetails`, etc. These are the elements contained in the container class and returned by the custom iterator (equivalent to `InTransaction` in the example above). These classes are very simple and contain only the fields necessary to do actual pagination: i.e. `symbol`, `since`, `limit`, `parameters`, etc. Perhaps these can be NamedTuples or dataclasses with frozen=true, to make sure they are read-only.
* add `def __iter__(self) -> "AbstractPaginationDetailsIterator"` method to `AbstractPaginationDetailSet` (it's not implemented in the abstract class but only in the subclasses)
* Add `DateBasedPaginationDetailsIterator` class. It's constructor receives all the information necessary to iterate: markets, current_results, etc. This class defines and updates iteration indexes
* Add `def __next__(self) -> DateBasedPaginationDetails:` method to `DateBasedPaginationDetailsIterator` which returns the next object and updates indexes. If there are no more objects it returns StopIteration

I think this will pay off in terms of design clarity: we want to make this superclass as clean as possible because it will be used by every CCXT plugin. 

I hope this all makes sense. Let me know if you have any questions or run into issues.

macanudo527:
The iterator is going to need to know what comes back from the exchange in order to return the next "element". So, we will need to do something like this:

```
pagination_detail_set: AbstractPaginationDetailSet = self.get_process_trades_pagination_detail_set(trades)
while pagination_details in pagination_detail_set:
            trades = self.__client.fetch_my_trades(
                symbol=pagination_details.symbol,
                since=pagination_details.since,
                limit=pagination_details.limit,
                params=pagination_details.parameters,
            )
        
     pagination_detail_set.updateIterator(trades) # This isn't possible right?
```

The iterator will need to know the live data coming back from the server in order to calculate the next date range. If there are more records and we hit the limit, we need to pull the window back. We will not know that information until we start pulling the data. So, is it possible to update the iterator in the middle of the iteration?

eprbell:
That's a really good point. I think we'll need to use the lower-level iterator API. Here's what I mean:
```
        pagination_detail_set: AbstractPaginationDetailSet = self.get_process_trades_pagination_detail_set()
        pagination_detail_iterator: AbstractPaginationDetailsIterator = iter(pagination_detail_set)
        try:
            while True:
            pagination_details = next(pagination_detail_iterator)
            trades = self.__client.fetch_my_trades(
                symbol=pagination_details.symbol,
                since=pagination_details.since,
                limit=pagination_details.limit,
                params=pagination_details.parameters,
            )
            pagination_detail_iterator.update_fetched_elements(trades)
        except StopIteration:
            # End of pagination details
            pass
```
Note that `AbstractPaginationDetailSet` and `AbstractPaginationDetailsIterator` remain as described in my previous comment, with the exception that `update_fetched_elements` needs to be added to `AbstractPaginationDetailsIterator`

Examples of this pattern in the code:
* use of `iter` to obtain an iterator object: https://github.com/eprbell/rp2/blob/main/src/rp2/tax_engine.py#L97
* use of `next` on the iterator and catching `StopIteration`: https://github.com/eprbell/rp2/blob/main/src/rp2/accounting_engine.py#L102

macanudo527:
It looks like the failed checks are due to a server outage.

eprbell:
We may want to mock test_plugin_ccxt.py to avoid these problems.

macanudo527:
I'm wondering what to name `symbol` in the pagination sets/iterators. Basically, we need to cycle through markets (eg. BTC/USD) for trades but codes/symbols (eg. BTC) of crypto assets for deposits/withdrawals. 

What would be a good clear name? Sector, grouping?

eprbell:
No strong opinion: perhaps symbol or product. We can also rename it later if we think of a better name.

macanudo527:
I have to do some cleaning, but core functionality should be available. I"m going to work on the implicit API and how to integrate that smoothly.

eprbell:
Hm, one more thing: it looks like CCXT may not be thread-safe (see https://github.com/ccxt/ccxt/issues/2252)

Two potential solutions:
* switch from thread pool to process pool. This may be a good idea regardless, because it reduces thread safety risks in any future CCXT clients we decide to support. Doing this is trivial: instead of `with ThreadPool(self.__thread_count) as pool:` use `with Pool(self.__thread_count) as pool:` (and update the import accordingly)
* nonce overriding approach (explained in the CCXT thread above)

Anyway, let's keep the parallel infrastructure in and keep exploring it: if we don't find any way of making CCXT work in parallel, we'll hardcode n_threads=1 (or remove the multithreading code), but I'm cautiously optimistic we can apply one of the above solutions successfully.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,900 @@
...
+                #   'currency': 'PAXG',
+                #   'status': 'ok',
+                #   'updated': None,
+                #   'internal': False,
+                #   'fee': None
+                # }
+
+                pagination_detail_iterator.update_fetched_elements(deposits)
+
+                with ThreadPool(self.__thread_count) as pool:
```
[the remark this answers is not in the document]
macanudo527: > I realized that this approach won't really get much speedup via multithreading, because it doesn't multithread the I/O-bound part of the code, which in this case would be fetch_deposits. Is there a way to call fetch_deposits on one single market? That way we could:

You can retrieve all deposits for a code (markets are for trades), but the efficiency of that approach would largely depend on the exchange. Most users wouldn't see a speed increase. Let's take Binance.com for example:

Let's say the user has 100 deposits in 10 different assets just for example's sake. 

Current approach:
1) Pull all deposits - Retrieve all 100 transactions at once. (1 point toward the rate limit)

Proposed approach:
1) fetch available codes (10 points)
2) Make 350+ calls, one for each of the possible assets held at Binance.com. Most of these will return nothing, but will still have to be checked. (350+ x 1 points)

I can see a speed advantage if a user runs an arbitrage bot that initiates many deposit/withdrawals with several different assets, but not with the average user. Binance.com actually allows a user to retrieve 1000 records in one call. I think that would cover most users in one call.

I think it will really depend on how the API on the exchange is structured. I'm afraid we don't have enough experience with the 100+ exchanges on CCXT between us to know the best approach. We might need to build what we can now and optimize later. 

I honestly don't see how multithreading could help withdrawals and deposits since they are relatively fast at the moment.  Trades, on Binance.com at least could theoretically be sped up with multithreading, but I'm already hitting the limit so multithreading would most likely not yield any speed improvements. 

Another thing to think about is that on some exchanges, CCXT may actually be pulling all the transactions and then sorting them by market/symbol and returning that, so multithreading would actually be much slower.

The only area that may have some small advantage is to pull trades, deposits, withdrawals, gains, and implicit API in separate threads. Or more likely, pull trades in one thread and everything else in another. This would speed things up for binance.com at least.

However, honestly, we really just don't know enough about the different structures to make a decision at this point. A lot of exchanges allow you to pull all transactions at once (like Coinbase), but others require you to cycle through markets/symbols. There are simply too many factors to really try to optimize right now. Can we just decide on a simple solution for now and move on? We may hash something out, only to have to rework the entire thing once we have more experience and realize something else might be much more efficient.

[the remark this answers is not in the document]
eprbell: That's fair: we can optimize as we get more experience. I have to admit my thinking on this was informed by my experience with the Coinbase and Coinbase Pro plugins (where multithreading improved performance 2-3x), but CCXT works differently and the example you gave is compelling. Let's revisit once we support more exchanges.

macanudo527:
Thinking about the implicit API. Do you think it would be better if the abstract class uses `dir()` to find implicit functions and then cycles through them. Something like this:
```
implicit_api_list = [method for method in dir(ExchangePlugin) if method.startswith('__') is False and method.startswith('_implicit_']
for implicit_function in method_list:
 # make I/O call
```

Each implicit API function would also need a custom `PaginationDetailSet` and a `post_processing()` function that would sort the json into the corresponding transactions after it is retrieved from the exchange, but these could be implied from the end of the API function. Given a function `_implicit_sapiGetAssetAssetDividend`, the abstract plugin would look for `_sapiGetAssetAssetDividendDetailSet` and `_sapiGetAssetAssetDividendPostProcessing` functions to match.

Do you think this is too complicated? or should there simply be a `_process_implicit_api()` function that must be overridden in the inherited class? and users can decide how to handle things.

I guess another option would be to have a `List` of implicit functions that the abstract plugin can iterate over and the iterator and post-processor can be implied from that.

eprbell:
The two approaches are not mutually exclusive: I was thinking we could do the simple approach first (`_process_implicit_api`), to get to MVP faster. Later maybe we can explore the other option and let the subclass decide which path to take.

macanudo527:
Ok, I added in gains. I made it a separate function from `_process_implicit_api()` to keep it better sorted. Also, I hope sometime in the future CCXT will implement it as one of their unified functions. I wrote a `CustomDatePaginationDetailSet` to handle the processing for these, but ended up not implementing it since I don't want to mess with the very delicate subscription/redemption calculation I have. I'd like to just get to an MVP as soon as I can.

macanudo527:
I just need to add dust trades, and then I should be able to wrap this up.

macanudo527:
OK, this is all wrapped up. Should I add something to the docs?

eprbell:
Great! I'll start reviewing. Yes, you can add reference to the abstract class to https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#data-loader-plugin-development

eprbell (changes requested):
Here's an initial review. Very exciting to see this take shape!

eprbell:
I've been out of town for a few days: I'll review in the weekend.

eprbell:
> OK, this is all wrapped up. Should I add something to the docs?

Sorry, forgot to answer this. Yes, we should have a section in the docs. Perhaps leave the existing section (which is still valid for exchanges unsupported by ccxt) and add a new one for ccxt.

eprbell (changes requested):
Here's another round. It's down to nitpicks mostly, but since this is going to be a core class used by all REST plugins I'd like to be extra thorough before merging.

macanudo527:
I'm  a little confused about what to make private/protected and how to access it in the subclass. 

Both `self.__logger` and `self.__client` are both needed in the subclasses in order to call the implicit API and log transactions. Should I keep the variables private and have `_get_logger()` and `_get_client()` functions so the subclasses can access them?

Or just have `self.__logger` and `self.__client` be protected variables? I'm not sure what the advantages of either approach is.

eprbell:
> I'm a little confused about what to make private/protected and how to access it in the subclass.
> 
> Both `self.__logger` and `self.__client` are both needed in the subclasses in order to call the implicit API and log transactions. Should I keep the variables private and have `_get_logger()` and `_get_client()` functions so the subclasses can access them?
> 
> Or just have `self.__logger` and `self.__client` be protected variables? I'm not sure what the advantages of either approach is.

Here's the basic idea for class members:
* private: visible only to the class, two leading underscores, no need to encapsulate it with a property or accessor function (i.e. `_get_stuff`)
*  protected: visible to the class and to its subclasses, one leading underscore, needs to be encapsulated with a property or an accessor function
* public: visible to everybody, no leading underscores, needs to be encapsulated with a property or acccessor function.

For public/protected members the choice of property vs accessor function is a matter of preference: I tend to use properties as simple accessors and accessor functions for anything more complex (or for things that don't exist in the class state and need to be built on the fly).

I hope this helps, let me know if you have more questions.

macanudo527:
> * protected: visible to the class and to its subclasses, one leading underscore, needs to be encapsulated with a property or an accessor function

So, the `logger` and `client` should just be protected right? I'm going to reuse them in the subclasses.

eprbell:
Yep, that sounds correct.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -178,69 +180,73 @@ def _process_deposits(
...
-                        intra_transactions.extend(processing_result.intra_transactions)
-
-        except StopIteration:
-            # End of pagination details
-            pass
+        pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_deposits_pagination_detail_set()
+        if not pagination_detail_set:
+            self._logger.error("No Pagination Details for Deposits")
+        else:
+            pagination_detail_iterator: AbstractPaginationDetailsIterator = iter(pagination_detail_set)
```
eprbell, line 187: Remove `else` to avoid unnecessary extra level of indentation:
```suggestion
        if not pagination_detail_set:
            self._logger.error("No Pagination Details for Deposits")
            return
        pagination_detail_iterator: AbstractPaginationDetailsIterator = iter(pagination_detail_set)
        ...
```

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -264,124 +270,130 @@ def _process_trades(
...
-                    if processing_result.out_transactions:
-                        out_transactions.extend(processing_result.out_transactions)
-
-        except StopIteration:
-            # End of pagination details
-            pass
+        pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_trades_pagination_detail_set()
+        if not pagination_detail_set:
+            self._logger.error("No Pagination Details for Trades")
+        else:
```
eprbell, line 276: Remove `else`

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -264,124 +270,130 @@ def _process_trades(
...
-                    if processing_result.intra_transactions:
-                        intra_transactions.extend(processing_result.intra_transactions)
-
-        except StopIteration:
-            # End of pagination details
-            pass
+        pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_withdrawals_pagination_detail_set()
+        if not pagination_detail_set:
+            self._logger.error("No Pagination Details for Withdrawals")
+        else:
```
eprbell, line 338: Remove `else`

eprbell (changes requested):
Minor changes.

macanudo527:
Could you get back to me on the questions above? re: `ProcessOperationResult`, `Trade`, `exchange.has[_FETCHMYTRADES], and `_to_trade`. And then, I think we are about finished.

macanudo527:
For `_process_gains()` and `_process_implicit_api()` should these just have `pass` in them in subclasses if the exchange doesn't require them? or have `pass` in the superclass and users can override the function if necessary?

eprbell:
> exchange.has[

Oh, sorry if I missed anything... I can't find any unresolved and unanswered questions in this PR (but Github can be a bit cumbersome with its folding of content): could you repeat the questions here?

eprbell:
> Huh, that's strange. It lists these all unresolved for me. Here they are:

I found the comments in file view (still not in the conversation view), but your reply isn't there: did you submit your comments? In any case I'll reply here.

> You commented about making `Trade` and `ProcessOperationResult` file-private:
> 
> > This should be file-private and have a leading underscore: `_ProcessOperationResult`
> 
> However, that would make them unavailable in the subclass right? I have the subclass setup to return results from single transaction processes with this tuple to support multi-threading.

True: moot point, then.

> I also use `_to_trade` in the subclasses to split apart the symbol into trade info that uses `Trade`
> 
> Also, my comment about renaming `_to_trade`:
> 
> I'm not that familiar with Python naming, but basically what this function is doing is parsing 3 variables into more useful information, so I think `_to` makes more sense. It is performing a conversion with the inputs not retrieving anything, which is what `_get` implies.

That's fine: it was just a suggestion and it's a matter of personal preference. Let's keep `_to`.

> My comments on `.has[]`:
> 
> > Should the check be inside _process_deposits? I'm thinking of the case of a subclass which may be adding functionality to _process_deposits: it would call super._process_deposits and then add its custom logic. However if the check is outside, the subclass custom logic will never be called.
> 
> > Same thing applies to the other checks in the lines below.
> 
> If `self.__client.has[_FETCH_DEPOSITS]` returns false the CCXT subclass does not have a `fetchDeposits` function, so the subclass would be calling an implicit API, which should be under `_process_implicit_api()`, right?

Another good point. Ok, let's leave it, then.

> I guess I saw these functions just covering the basic unified API and anything else needs to be covered in `_process_implicit_api()` or `_process_gains()`. What would be the use case for overriding `_process_deposits`?

Yeah, I was thinking the user might add implicit "deposit" APIs to their own version of `_process_deposit()`, but I think you're right: they would just add them to `_process_implicit_api()`, which is a simpler model.

> Also, I don't think it is a good idea to encourage these to be overwritten since it might make mult-threading more complicated. If we know these only cover the unified API it will be easier to optimize and prevent race conditions.

So it sounds like we can ignore all of these comments and move on. I'll review everything once more and then we can merge.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,628 @@
...
+_SYMBOL: str = "symbol"
+_TIMESTAMP: str = "timestamp"
+_TYPE: str = "type"
+_TX_ID: str = "txid"  # CCXT doesn't capitalize I
+_WITHDRAWAL: str = "withdrawal"
+
+_MS_IN_SECOND: int = 1000
+
+
+class Trade(NamedTuple):
```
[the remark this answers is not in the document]
eprbell: Ignore: resolved in separate conversation in main thread.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,628 @@
...
+    base_asset: str
+    quote_asset: str
+    base_info: str
+    quote_info: str
+
+
+class ProcessOperationResult(NamedTuple):
```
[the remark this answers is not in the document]
eprbell: Ignore: resolved in separate conversation in main thread.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,628 @@
...
+        return self.__markets
+
+    @staticmethod
+    def _rp2_timestamp_from_ms_epoch(epoch_timestamp: str) -> str:
+        rp2_time = datetime.fromtimestamp((int(epoch_timestamp) / _MS_IN_SECOND), timezone.utc)
+
+        return rp2_time.strftime("%Y-%m-%d %H:%M:%S%z")
+
+    @staticmethod
+    def _to_trade(market_pair: str, base_amount: str, quote_amount: str) -> Trade:
```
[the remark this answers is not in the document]
eprbell: We'll leave it as `_to_trade`: resolved in separate conversation in main thread.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,628 @@
...
+            quote_info=f"{quote_amount} {assets[1]}",
+        )
+
+    def load(self) -> List[AbstractTransaction]:
+        result: List[AbstractTransaction] = []
+        in_transactions: List[InTransaction] = []
+        out_transactions: List[OutTransaction] = []
+        intra_transactions: List[IntraTransaction] = []
+
+        if self._client.has[_FETCH_DEPOSITS]:
```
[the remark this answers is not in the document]
eprbell: Ignore: resolved in separate conversation in main thread.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+    ### Multiple Transaction Processing or Pagination Methods
+
+    def _process_deposits(
+        self,
+        intra_transactions: List[IntraTransaction],
+    ) -> None:
+
+        pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_deposits_pagination_detail_set()
+        # Strip optionality
+        if not pagination_detail_set:
```
eprbell, line 185: Shouldn't we throw an exception here, instead of logging? If `pagination_detail_set` is `None` calling `iter` on it a few lines down will cause a crash.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+            self._logger.error("No Pagination Details for Deposits")
+        else:
```
eprbell, line 187: After adding the exception, remove the else and reduce indentation level of the next line.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+        out_transactions: List[OutTransaction],
+    ) -> None:
+
+        processing_result_list: List[Optional[ProcessOperationResult]] = []
+        pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_trades_pagination_detail_set()
+        # Strip optionality
+        if not pagination_detail_set:
+            self._logger.error("No Pagination Details for Trades")
+        else:
+            has_pagination_detail_set: AbstractPaginationDetailSet = pagination_detail_set
```
eprbell, line 281: Same as above: use exception instead of logger and remove else.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,628 @@
...
+        out_transactions: List[OutTransaction],
+    ) -> None:
+
+        processing_result_list: List[Optional[ProcessOperationResult]] = []
+        pagination_detail_set: AbstractPaginationDetailSet = self._get_process_trades_pagination_detail_set()
+        pagination_detail_iterator: AbstractPaginationDetailsIterator = iter(pagination_detail_set)
+        try:
+            while True:
+                pagination_details: PaginationDetails = next(pagination_detail_iterator)
+                trades: Optional[List[Dict[str, Union[str, float]]]] = self._client.fetch_my_trades(
```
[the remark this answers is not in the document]
eprbell: Any thoughts on this?

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+        self,
+        intra_transactions: List[IntraTransaction],
+    ) -> None:
+
+        pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_withdrawals_pagination_detail_set()
+        # Strip optionality
+        if not pagination_detail_set:
+            self._logger.error("No Pagination Details for Withdrawals")
+        else:
+            has_pagination_detail_set: AbstractPaginationDetailSet = pagination_detail_set
```
eprbell, line 346: Same as above: use exception instead of logger and remove else.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+            crypto_in = RP2Decimal(str(transaction[_COST]))
+            conversion_info = f"{trade.base_info} -> {trade.quote_info}"
+        else:
+            raise Exception(f"Internal error: unrecognized transaction side: {transaction[_SIDE]}")
+
+        if fee_asset == in_asset:
+            crypto_fee = RP2Decimal(str(transaction[_FEE][_COST]))
+        else:
+            crypto_fee = ZERO
+            transaction_fee = RP2Decimal(transaction[_FEE][_COST])
```
eprbell, line 442: Is transaction[_FEE][_COST] a float? If so the right way to convert it is: `RP2Decimal(str(transaction[_FEE][_COST]))`

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+        if transaction[_FEE][_CURRENCY] == out_asset:
+            crypto_fee: RP2Decimal = RP2Decimal(str(transaction[_FEE][_COST]))
+        else:
+            crypto_fee = ZERO
+        crypto_out_with_fee: RP2Decimal = crypto_out_no_fee + crypto_fee
+
+        # Is this a plain buy or a conversion?
+        if self.is_native_fiat(trade.quote_asset):
+            fiat_out_no_fee: RP2Decimal = RP2Decimal(str(transaction[_COST]))
+            fiat_fee: RP2Decimal = RP2Decimal(crypto_fee)
```
eprbell, line 552: Seems to me `crypto_fee` is a RP2Decimal already. Can we skip the conversion?

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,651 @@
...
+        exchange_start_time: datetime,
+        native_fiat: str,
+        thread_count: Optional[int],
+    ) -> None:
+
+        super().__init__(account_holder, native_fiat)
+        self._logger: logging.Logger = create_logger(f"{self.exchange_name()}/{self.account_holder}")
+        self.__cache_key: str = f"{str(self.exchange_name).lower()}-{account_holder}"
+        self._client: Exchange = self._initialize_client()
+        self._thread_count = thread_count if thread_count else self.__DEFAULT_THREAD_COUNT
```
eprbell, line 96: All fields still need to start with double underscore (our data models are encapsulated and we never access fields directly outside the class in which they are defined). To implement protected behavior they need a property that starts with single underscore (to access in the subclass). Sorry if my description wasn't clear when we discussed this.

Here's an updated text that's hopefully clearer:
* all class fields have two leading underscores because our data models are encapsulated and we never access fields directly outside the class in which they are defined
* private fields: visible only to the class, nothing else to do (access within the class via the double-underscore filed)
* protected fields: visible to the class and to its subclasses, needs to be encapsulated with a property or an accessor function that starts with one leading underscore 
* public fields: visible to everybody, needs to be encapsulated with a property or acccessor function that starts with no leading underscore.

eprbell (changes requested):
A few minor fixes needed.

eprbell:
Unfortunately I cannot :-( 

When you look at your comments do they show up as "pending"? If so you need to submit them to make them publicly visible. Not sure what else this could be, short of a site malfunction...

macanudo527 (commented):
Ok, now, you should see all comments, right?

macanudo527:
Ok, now, everything should show. I have no idea how I have used Github for so long and not realized you have to submit all your comments. I have no idea what I was doing before.

Sorry about that.

eprbell:
Now I see it: mystery solved :-) Will take a look in the next 2-3 days (currently traveling).

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,878 @@
...
+        markets: Optional[List[str]] = None,
+        params: Optional[Dict[str, Union[int, str, None]]] = None,
+        window: Optional[int] = None,
+    ) -> None:
+
+        super().__init__(limit, markets, params)
+        self.__end_of_data = False
+        self.__since: Optional[int] = exchange_start_time
+        self.__exchange_start_time: int = exchange_start_time
+        self.__now: int = int(datetime.now().timestamp()) * _MS_IN_SECOND
```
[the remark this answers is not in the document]
eprbell: Yeah, on second thought I think `now` has fairly clear meaning. Let's leave it as-is.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,878 @@
...
+            return _PaginationDetails(
+                symbol=self.get_market(),
+                since=self.__since,
+                limit=self.limit,
+                params=self.params,
+            )
+        raise StopIteration(self)
+
+    @property
+    def window(self) -> int:
```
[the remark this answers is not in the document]
eprbell: Let's rename the property as `_window` to make it protected. Note that all fields start with two underscores, but to give them protected access we add a single underscore property (and for public access we add a no-underscore property).

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,878 @@
...
+    def __init__(
+        self,
+        exchange_start_time: int,
+        start_time_key: str,
+        end_time_key: str,
+        window: int,
+        limit: Optional[int] = None,
+        markets: Optional[List[str]] = None,
+        params: Optional[Dict[str, Union[int, str, None]]] = None,
+    ) -> None:
```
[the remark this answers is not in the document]
eprbell: Same as comment above

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,878 @@
...
+
+        super().__init__(account_holder, native_fiat)
+        self.__logger: logging.Logger = create_logger(f"{self.exchange_name()}/{self.account_holder}")
+        self.__cache_key: str = f"{str(self.exchange_name).lower()}-{account_holder}"
+        self.__client: Exchange = self.initialize_client()
+        if native_fiat:
+            self.__native_fiat: str = native_fiat
+        else:
+            self.__native_fiat = "USD"
```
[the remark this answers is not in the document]
eprbell: Actually you're right, we shouldn't do anything to `native_fiat` and simply let the superclass deal with it. So just remove these 4 lines and leave it optional. Does that work?

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,878 @@
...
+    def native_fiat(self) -> str:
+        return self.__native_fiat
+
+    @property
+    def start_time_ms(self) -> int:
+        return self.__start_time_ms
+
+    @property
+    def thread_count(self) -> int:
+        return self.__thread_count
```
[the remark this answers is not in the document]
eprbell: The accessor should not be duplicated here since it's already in the superclass: we can remove it from this class. But you're right about it being defined as public in `AbstractInputPlugin`. Don't worry about that: I'll take a look at it and adjust the scope.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -0,0 +1,628 @@
...
+                #       'amount':       1.5,                        // amount of base currency
+                #       'cost':         0.10376526,                 // total cost, `price * amount`,
+                #       'fee':          {                           // provided by exchange or calculated by ccxt
+                #           'cost':  0.0015,                        // float
+                #           'currency': 'ETH',                      // usually base currency for buys, quote currency for sells
+                #           'rate': 0.002,                          // the fee rate (if available)
+                #       },
+                #   }
+
+                # * The work on ``'fee'`` info is still in progress, fee info may be missing partially or entirely, depending on the exchange capabilities.
```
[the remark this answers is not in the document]
eprbell: I think it may be misleading to paste a direct quote from their docs, because their software may change and then our comment would be out of sync. How about a link to their doc or nothing at all?

eprbell (commented):
OK, I think I replied to all of the comments. Let me know if anything is still missing.

src/dali/abstract_ccxt_input_plugin.py
eprbell: So, actual fields have leading double underscore:
```
self.__client
self.__thread_count
```

Then:
* if we want private access (this class only) we do nothing else
* if we want protected access (this class and its subclasses), we add a property with a single underscore (or an accessor method starting with `_get`)
* if we want public access (everybody), we add a property with no underscore (or an accessor method starting with `get`)

Whether to use a property or a get function is up to you: I tend to use properties if the access logic is fairly simple and get functions if access logic is more complicated. But again, either way works. Hope this is clearer, sorry if I didn't articulate it clearly enough earlier.

src/dali/abstract_ccxt_input_plugin.py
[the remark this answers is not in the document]
eprbell: Ok, sounds good.

eprbell (commented):
More answers.

src/dali/abstract_ccxt_input_plugin.py
eprbell: I added a data encapsulation paragraph (with the above description) to the Design Guidelines section of README.dev.md both in RP2 and DaLI. Let me know if you think it needs more explanation: I'd like to fix the docs so that nobody trips on it in the future.

macanudo527:
I think I more or less wrapped everything up. I overrode the `_client` property to type it for mypy. Let me know if that is the right way to do it.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -341,9 +356,9 @@ def _process_withdrawals(
         pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_withdrawals_pagination_detail_set()
         # Strip optionality
         if not pagination_detail_set:
-            self._logger.error("No Pagination Details for Withdrawals")
-        else:
-            has_pagination_detail_set: AbstractPaginationDetailSet = pagination_detail_set
+            raise Exception("No Pagination Details for Deposits")
```
eprbell, line 359: It should be withdrawals, not deposits.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -276,9 +292,9 @@ def _process_trades(
         pagination_detail_set: Optional[AbstractPaginationDetailSet] = self._get_process_trades_pagination_detail_set()
         # Strip optionality
         if not pagination_detail_set:
-            self._logger.error("No Pagination Details for Trades")
-        else:
-            has_pagination_detail_set: AbstractPaginationDetailSet = pagination_detail_set
+            raise Exception("No Pagination Details for Deposits")
```
eprbell, line 295: Should this be deposits or trades?

src/dali/plugin/input/rest/binance_com.py
```diff
@@ -186,11 +186,18 @@ def _initialize_client(self) -> binance:
             }
         )
 
+    @property
+    def _client(self) -> binance:
```
eprbell, line 190: Not sure about changing the return type: I guess if mypy doesn't complain I'm good with it (the other option would have been to leave the return type as `Exchange` and use `cast` in the call sites). Otherwise it's fine to override the property because you're adding extra runtime checks.

eprbell (changes requested):
Minor typos. I'll do one more full review once those are fixed and then we'll merge.

eprbell (approved):
LGTM!

eprbell:
Grats! This is an important merge.

macanudo527:
It's good to see this go through! Thanks for all the help on the differences with protected / private.

PR #84, opened 2022-09-13
Pionex plugin

macanudo527:
This is a CSV plugin for the Pionex Exchange. They did just recently implement a REST API, but CCXT hasn't adopted it yet. This is meant as a stop-gap until it does.


eprbell (approved):
LGTM.

PR #86, opened 2022-09-21
Bitbank REST Plugin

macanudo527:
The REST API for this exchange only supports `fetch_my_trades`. I'll submit another PR for the supplemental CSV plugin that covers deposits and withdrawals.

This PR is dependent on #80 and shouldn't be approved until that one is.


src/dali/plugin/input/rest/bitbank.py
```diff
@@ -0,0 +1,105 @@
...
+        return bitbank(
+            {
+                "apiKey": self.__api_key,
+                "enableRateLimit": True,
+                "secret": self.__api_secret,
+            }
+        )
+
+    def _get_process_deposits_pagination_detail_set(self) -> AbstractPaginationDetailSet:
+        pass
```
eprbell, line 80: I'm surprised this passes static analysis: this function expects a pagination detail set object as return value. I think we should change this method (here and in the superclass) to return `Optional[AbstractPaginationDetailSet]` so that this method can simply return None.

eprbell (changes requested):
Amazing, how easy it is to define a new REST loader now!

eprbell:
Also, add docs for the two Bitbank plugins.

macanudo527:
I'm going to add docs for the Abstract CCXT plugin, bitbank, pionex, and coincheck plugins in a separate PR.

eprbell:
After we merged #80. this one has a few conflicts that need fixing. Then I'll review and we can merge this as well.

macanudo527:
This should be good now.

eprbell (approved):
I still can't believe how easy it is to add a new data loader plugin now!

macanudo527:
Well, this is a pretty thin plugin, but yeah, basic functionality is possible with a relatively short (and simple) piece of code.

eprbell:
> Well, this is a pretty thin plugin, but yeah, basic functionality is possible with a relatively short (and simple) piece of code.

True, but without the abstract superclass it would be a lot more code :-)

PR #87, opened 2022-09-22
Bitbank Supplemental CSV

macanudo527:
This allows you to process the CSV file for crypto withdrawals on Bitbank. I haven't made any deposits, so I don't know the format. 

Fiat deposits have no fees.


src/dali/plugin/input/csv/bitbank_supplemental.py
```diff
@@ -0,0 +1,97 @@
...
+        with open(file_path, encoding="utf-8") as csv_file:
+            lines = reader(csv_file)
+
+            header = next(lines)
+            self.__logger.debug("Header: %s", header)
+            for line in lines:
+                raw_data: str = self.__DELIMITER.join(line)
+                self.__logger.debug("Transaction: %s", raw_data)
+
+                jst_timezone = PytzTimezone("Asia/Tokyo")
```
eprbell, line 76: Is this safe: i.e. Is this a Japan-only exchange?

src/dali/plugin/input/csv/bitbank_supplemental.py
```diff
@@ -0,0 +1,97 @@
...
+        self,
+        account_holder: str,
+        withdrawals_csv_file: str,
+        withdrawals_code: str,
+        native_fiat: Optional[str] = None,
+    ) -> None:
+
+        super().__init__(account_holder=account_holder, native_fiat=native_fiat)
+        self.__withdrawals_csv_file: str = withdrawals_csv_file
+        self.__withdrawals_code: str = withdrawals_code
```
eprbell, line 58: Could you add a small comment describing what a withdrawal code is?

eprbell (changes requested):
Just a few nits.

src/dali/plugin/input/csv/bitbank_supplemental.py
macanudo527: Yes, it"s Japan only. Japan has similar (but less restrictive) exchange rules. Any exchange that deals in JPY fiat must be registered in Japan and only in Japan. Of course, users are still free to transfer crypto off to another exchange that doesn't have such restrictions. These transactions have to be labeled though according to the "Travel Rule", so the Japanese government has a general idea of the flow of the assets.

This means a lot of the major exchanges have a special Japanese version (e.g. Coinbase has Coinbase Japan that operates independently from Coinbase US)

macanudo527: I should also add that there is no timezone stamp, so we have to assume some time zone, unfortunately.

macanudo527:
This should be wrapped up as well.

src/dali/plugin/input/csv/bitbank_supplemental.py
eprbell: Thanks for the explanation. The Trezor plugin had a similar problem (no timezone), so it lets the user specify it in the constructor: https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/csv/trezor.py#L52. 

There are two options: 
- the exchange assumes the timezone to be `Asia/Tokyo`: in this case there is nothing else to do
- the exchange records time with the user's local timezone as the implied timezone: in this case the Trezor approach can be useful (maybe have `Asia/Tokyo` as default value).

Do we know which of the two options applies here?

macanudo527: That is a good question, but I'm going to guess that it will always be `Asia/Tokyo` at least while this exchange operates exclusively in Japan. Japan only has one timezone, so even if it was based on user's local timezone, it would probably not matter that much. Only Japanese residents can register for the exchange, so in order for the user's local timezone and server timezones to be different, it would have to be a Japanese resident travel overseas at the moment of the transaction. That does occur, but probably pretty rarely. 

Anyway, I can't seem to find any documentation on it. Their site doesn't contain any information as to whether it is definitively server time or local user's time. The REST API uses unix timestamps so doesn't have this issue at all.

eprbell (approved):
LGTM

PR #88, opened 2022-09-25
Coincheck Supplemental plugin

macanudo527:
Coincheck only really has an exchange for BTC at the moment. Everything else is sold on a marketplace which is not part of the REST API. You can't even retrieve transfers via the REST API, so this plugin just processes marketplace buys. It's incomplete, but I'm not sure how much longer Coincheck will be in business. They suffered a major attack a couple of years ago and haven't really recovered.

I no longer use them. This is just for a few minor old buys I have.


tests/test_plugin_coincheck.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2022 macanudo527
```
eprbell, line 1: Is this file a duplicate of tests/test_plugin_coincheck_supplemental.py?

eprbell (changes requested):
We'll need to remove spurious files (binance, etc.) before merging. Also add documentation for this plugin.

macanudo527:
At this point, I think it would just be easier to wait for #80 to get merged, and then I'll patch up these other pulls.

macanudo527:
With the merge, this should be clean now.

src/dali/plugin/input/csv/coincheck_supplemental.py
```diff
@@ -0,0 +1,94 @@
...
+            lines = reader(csv_file)
+
+            header = next(lines)
+            self.__logger.debug("Header: %s", header)
+            for line in lines:
+                raw_data: str = self.__DELIMITER.join(line)
+                self.__logger.debug("Transaction: %s", raw_data)
+
+                unformatted_timestamp: str = line[self.__TIMESTAMP_INDEX]
+                timestamp: str = unformatted_timestamp[: -len(" UTC")] if unformatted_timestamp.endswith(" UTC") else unformatted_timestamp
```
eprbell, line 73: Curious about this logic: do some timestamps end with UTC and some not?

src/dali/plugin/input/csv/coincheck_supplemental.py
```diff
@@ -0,0 +1,94 @@
...
+
+                result.append(
+                    InTransaction(
```
eprbell, line 76: This InTransaction is a buy with no fee (either fiat or crypto): is this correct? I know exchanges sometimes offer zero-fee buys but in my experience they are limited in number and when they're over fees start appearing.

src/dali/plugin/input/csv/coincheck_supplemental.py
macanudo527: All timestamps that I have seen end with UTC. I guess I just added that for redundancy. Should I take it out?

src/dali/plugin/input/csv/coincheck_supplemental.py
macanudo527: This is a marketplace buy and not a trade. This exchange sells crypto assets the same way you would exchange money at like Travelex. They have a wide spread ~5% and one price for buys and one for sells.

Coincheck only has a trading 'floor' for BTCJPY.

src/dali/plugin/input/csv/coincheck_supplemental.py
eprbell: I would suggest taking it out, if there is no use case for it.

src/dali/plugin/input/csv/coincheck_supplemental.py
eprbell: Thanks for the explanation: sounds good, then.

src/dali/plugin/input/csv/coincheck_supplemental.py
macanudo527: OK, I took it out.

eprbell (approved):
LGTM

PR #98, opened 2022-10-26
Ccxt fixes

macanudo527:
There are a lot of very minor changes in this push. Basically, there are a handful of missing pieces I put in.

This is not complete yet. It might be 2 or 3 days before I can get a full run in. I just wanted to check with you on replacing the `None` lists with empty lists. I'm not sure if this is the right call or if it should be done earlier in the process.


src/dali/ccxt_pagination.py
```diff
@@ -16,6 +16,7 @@
 # https://docs.ccxt.com/en/latest/index.html
 
 # pylint: disable=fixme
+# pylint: disable=dangerous-default-value
```
eprbell, line 19: Disabling this is a bit dangerous: see comment below.

src/dali/ccxt_pagination.py
```diff
@@ -49,7 +50,7 @@ def __init__(
         exchange_start_time: int,
         limit: Optional[int] = None,
         markets: Optional[List[str]] = None,
-        params: Optional[Dict[str, Union[int, str, None]]] = None,
+        params: Optional[Dict[str, Union[int, str, None]]] = {},
```
eprbell, line 53: Using an empty dictionary as default value can be problematic, see: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

Using `None` is safe and customary in cases like this: what was the rationale for the change?

eprbell (commented):
Initial feedback.

src/dali/ccxt_pagination.py
macanudo527: CCXT  needs a list for `params` or no `params`. It will throw an error if you pass `None`. So, basically, `params` should never be `None`.

eprbell: Got it: in that case you can leave the parameter default as `None` and normalize `params` at the beginning of the function:
```
params = {} if None else params
```

src/dali/ccxt_pagination.py
```diff
@@ -117,16 +115,17 @@ def __iter__(self) -> "CustomDateBasedPaginationDetailsIterator":
 
 class AbstractPaginationDetailsIterator:
     def __init__(self, limit: Optional[int], markets: Optional[List[str]] = None, params: Optional[Dict[str, Union[int, str, None]]] = None) -> None:
+        params = {} if params is None else params
         self.__limit: Optional[int] = limit
-        self.__markets: Optional[List[str]] = markets
+        self.__markets: Optional[List[str]] =  markets
```
eprbell, line 120: Unnecessary white space.

src/dali/ccxt_pagination.py
```diff
@@ -162,28 +161,30 @@ def __init__(
...
-        self.__window: int = window if window else _DEFAULT_WINDOW
+        self.__window: int = window
 
     def update_fetched_elements(self, current_results: Any) -> None:
 
         end_of_market: bool = False
 
         # Update Since if needed otherwise end_of_market
-        if len(current_results):
+        if len(current_results) == self._get_limit:
```
eprbell, line 171: Should this be:
```suggestion
        if len(current_results) == self._get_limit():
```

src/dali/plugin/input/rest/bitbank.py
```diff
@@ -103,3 +117,34 @@ def _process_implicit_api(
...
     ) -> None:
         pass
+
+    # Override the handling of trades since Bitbank has negative maker fees
+    def _process_buy_and_sell(self, transaction: Any, notes: Optional[str] = None) -> ProcessOperationResult:
+        fee_income: Optional[InTransaction] = None
+        if RP2Decimal(str(transaction[_FEE][_COST])) < ZERO:
+            fee_income = InTransaction(
+                    plugin=self.plugin_name(),
+                    unique_id=f"FI{transaction[_ID]}",
```
eprbell, line 127: Why the prepended "FI": wouldn't it throw off the transaction resolver if we mangle the unique id?

eprbell (commented):
Just a few comments.

src/dali/plugin/input/rest/bitbank.py
macanudo527: I guess for some reason I thought it would confuse the transaction resolver. I guess 'fees' like this won't bother it.

macanudo527:
This should be clean. I'm running another full pull now to be sure everything works as expected.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -420,6 +439,48 @@ def _process_withdrawals(
...
+        # Most exceptions are caused by request limits of the underlying APIs
+        while request_count < 9:
+            try:
+                if "code" in params:
+                    results = function(
+                        code=params["code"],
+                        since=params["since"],
+                        limit=params["limit"],
+                        params=params["params"],
+                    )
```
eprbell, line 460: You should be able to use dict unpacking here:
```suggestion
                    results = function(**params)
```

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -420,6 +439,48 @@ def _process_withdrawals(
...
+                else:
+                    results = function(
+                        symbol=params["symbol"],
+                        since=params["since"],
+                        limit=params["limit"],
+                        params=params["params"],
+                    )
```
eprbell, line 467: Dict unpacking

src/dali/dali_main.py
```diff
@@ -162,6 +162,9 @@ def _dali_main_internal(country: AbstractCountry) -> None:
         with ThreadPool(args.thread_count) as pool:
             result_list = pool.map(_input_plugin_helper, input_plugin_args_list)
 
+        # prune None lists from the result_list
+        result_list = [[] if plugin_results is None else plugin_results for plugin_results in result_list]
```
eprbell, line 166: Is this necessary? The following line (extracting transactions) should work even with empty lists.

eprbell (changes requested):
A couple of small comments, otherwise LGTM.

src/dali/abstract_ccxt_input_plugin.py
macanudo527: That's convenient. Learn something new every day. Thanks.

src/dali/dali_main.py
macanudo527: There was a `NoneType` that kept popping up that I couldn't track down that was crashing the list comprehension in the next line, but I must have fixed it with the latest bug squashes. I took this line out.

macanudo527:
I made a full pull and everything came up aces. This is good to go with real-world data.

eprbell (approved):
LGTM.

PR #108, opened 2023-01-11
Fix Fiat deposits

macanudo527:
This is mostly a fix to add fiat deposits no matter what they are (native or non-native), which is something I overlooked when I first coded it. 

I still need to add documentation to Bitbank supplemental plugin for this.

There is also a minor fix for filtering out zero'd entries from Pionex.

I need to compute, file taxes, and work out any bugs involved with that for now. I'll try to double back and refactor the RP2 JP report code once I'm finished.


src/dali/transaction_resolver.py
```diff
@@ -231,62 +231,72 @@ def resolve_transactions(
     unique_id_2_transactions: Dict[AssetAndUniqueId, List[AbstractTransaction]] = {}
     transaction: AbstractTransaction
 
-    for transaction in transactions:
```
eprbell, line 234: AFAICT the main addition here is the try catch, but the combination of new indentation + formatting changes (like in line 250/251) throws off the diff algorithm and we get a large diff as a result.

Since the formatting changes are not relevant to this PR and they cause the diff to be unnecessarily large, could you undo the formatting changes and only leave in the logic you want to change? If you have formatting changes to propose it's probably better to do it in a different PR. 

The reason I'm more conservative than usual with changes to this file is that the transaction resolver is a central element of the system and I want to make sure I review all logic changes thoroughly (and mixing in formatting changes makes that a bit harder).

eprbell (changes requested):
Initial review.

src/dali/transaction_resolver.py
macanudo527: This is a major issue for me. I need to interrupt the process and save the cache file. Basically, it still takes 10-12 hours to make a full pull on my laptop which I need to unplug and move with me. Saving the cache file now takes a significant amount of time so it is often interrupted if I save periodically. I need to save on exit in order to test.

Since I'm under a deadline (I need to file taxes). I'll have to unfortunately keep this in place for now while I finish working. I'll try to remember to roll it back before submitting.

eprbell: To clarify further: I am not requesting to roll back the new functionality (saving the cache), but only the formatting changes (such as those on line 250/251). The reason is that new indentation + formatting changes seems to throw off the diff algorithm which generated a large diff unnecessarily (and that is harder to review). I do understand you're on a deadline though: there is no rush.

macanudo527:
I'm having issue getting the `console_scripts` to work.  I added `dali_jp` to `setup.cfg`, but it doesn't recognize the command. I'm sure this is a stupid question, but how do I build the binary?

eprbell:
> 

Can you share the exact error you're getting? The procedure should be:
* add a new jp country plugin in DaLI (see dali/plugin/country/us.py for an example)
* add a new dali_jp line in the console_scripts section of setup.cfg

When you install DaLI it installs a local copy of rp2: perhaps that copy is stale? You can try to make clean and then reinstall DaLI.

macanudo527:
That (reinstall dali) did it. Thanks! I'm still rusty with general programming building/installing.

src/dali/abstract_ccxt_input_plugin.py
```diff
@@ -544,7 +544,7 @@ def _process_buy(self, transaction: Any, notes: Optional[str] = None) -> Process
             in_transaction_list.append(
                 InTransaction(
                     plugin=self.plugin_name(),
-                    unique_id=transaction[_ID],
+                    unique_id=f"B{transaction[_ID]}",
```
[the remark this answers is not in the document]
macanudo527: I was having issues with the transaction resolver linking the buy and sell transactions of JPY into transfers for some reason.

For Binance, the purchases are  by credit card. One Binance transaction becomes 3 in dali:

> 1) InTransaction for Deposit of JPY
> 2) OutTransaction for conversion of JPY (since I'm filing US taxes, this isn't a buy)
> 3) InTransaction of crypto asset

If I remember correctly, the transaction resolver combined 1) and 2) and interpreted it as a transfer.

The other case was I think with Bitbank, which PAYS maker fees (i.e. income for providing liquidity), a trade on that site also creates 3 transactions:

> a) OutTransaction for JPY purchases
> b) InTransaction of JPY income for providing liquidity
> c) InTransaction of crypto asset

In this case, the transaction resolver combined a) and b) and interpreted it as a transfer.

I guess I can just mangle the id for the Binance fiat deposit 1) and the id for the income from Bitbank b).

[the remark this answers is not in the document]
eprbell: This is making me realize that the transaction resolver should operate per-asset, so that no mangling would be necessary in a case like this, but instead it operates globally. So I would say: mangle the fiat transactions by appending `"/fiat_buy"` or `"/fiat_sell"`. Also maybe add comments describing these artificial fiat transactions.

macanudo527:
Is it all right if we just add a check at [line #268 after the two transactions](https://github.com/eprbell/dali-rp2/blob/85621d3cc8141a051f3257e753657202edb3adef/src/dali/transaction_resolver.py#L268) are assigned to see if both transactions have the same exchange? Like this:

```
if transaction1.exchange == transaction2.exchange:
    continue
```

Is there a problem with that? Transfers will always be between two different exchanges, right?

eprbell:
No, the exchange field could be the same. E.g. two people filing together could have two separate accounts on binance and move coins between the two accounts.

macanudo527:
Ok, how about at [line#595](https://github.com/eprbell/dali-rp2/blob/85621d3cc8141a051f3257e753657202edb3adef/src/dali/transaction_resolver.py#L595), we check holders:

```
if from_exchange == to_exchange and from_holder == to_holder:
    return None
```

Of course, this means you have to change the return signature and check for `None`. It's a little sloppy I guess.

src/dali/transaction_resolver.py
macanudo527: OK, these formatting changes were because the line was too long. Should I just disable that check for now?

eprbell:
> Ok, how about at [line#595](https://github.com/eprbell/dali-rp2/blob/85621d3cc8141a051f3257e753657202edb3adef/src/dali/transaction_resolver.py#L595), we check holders:
> 
> ```
> if from_exchange == to_exchange and from_holder == to_holder:
>     return None
> ```
> 
> Of course, this means you have to change the return signature and check for `None`. It's a little sloppy I guess.

If there is no major advantage to doing so, I would leave the logic unchanged (it seems a corner case to me).

src/dali/plugin/input/csv/bitbank_supplemental.py
```diff
@@ -48,22 +51,81 @@ class InputPlugin(AbstractInputPlugin):
...
+            result.extend(self.parse_deposits_file(self.__deposits_csv_file))
+
+        return result
+
+    def parse_deposits_file(self, file_path: str) -> List[AbstractTransaction]:
+        result: List[AbstractTransaction] = []
+        if self.__deposits_code:
+            deposits_code: str = self.__deposits_code
+        else:
+            raise Exception("Bitbank.cc supplemental plugin deposits file declared without deposits code.")
```
eprbell, line 85: Raising Exception is reserved for internal errors (or sanity errors). In the case of user errors, you can use `RP2ValueError` or `RP2TypeError`.

src/dali/plugin/input/csv/bitbank_supplemental.py
```diff
@@ -48,22 +51,81 @@ class InputPlugin(AbstractInputPlugin):
...
+                    )
+
+        return result
 
     def parse_withdrawals_file(self, file_path: str) -> List[AbstractTransaction]:
         result: List[AbstractTransaction] = []
+        if self.__withdrawals_code:
+            withdrawals_code: str = self.__withdrawals_code
+        else:
+            raise Exception("Bitbank.cc supplemental plugin withdrawals file declared without withdrawals code.")
```
eprbell, line 128: Same as above: use `RP2ValueError` or `RP2TypeError`.

src/dali/plugin/input/csv/bitbank_supplemental.py
```diff
@@ -48,22 +51,81 @@ class InputPlugin(AbstractInputPlugin):
...
+                            notes="Fiat deposit",
+                        )
+                    )
+
+        return result
 
     def parse_withdrawals_file(self, file_path: str) -> List[AbstractTransaction]:
         result: List[AbstractTransaction] = []
+        if self.__withdrawals_code:
+            withdrawals_code: str = self.__withdrawals_code
```
eprbell, line 126: Just a nit, but this new variable seems unnecessary (just access the instance variable)

src/dali/transaction_resolver.py
eprbell: No problem: on second thought it's better to just follow the indications of static analysis without suppressing anything. In the end I diffed this file using Emacs and that diff was very easy to read.

src/dali/transaction_resolver.py
```diff
@@ -231,62 +231,72 @@ def resolve_transactions(
...
+            LOGGER.debug("Resolved transaction: %s", str(transaction))
+            resolved_transactions.append(transaction)
 
-        LOGGER.debug("Resolved transaction: %s", str(transaction))
-        resolved_transactions.append(transaction)
+    except KeyboardInterrupt as exc:
+        LOGGER.info("Exiting and saving to cache.")
+        for pair_converter in global_configuration[Keyword.HISTORICAL_PAIR_CONVERTERS.value]:
+            cast(AbstractPairConverterPlugin, pair_converter).save_historical_price_cache()
+        raise Exception("Dali terminated by user.") from exc
```
eprbell, line 299: Raising `Exception` is reserved for internal / sanity errors. In this case I would just rethrow the same exception without modification by using `raise` alone.
``` suggestion
        raise
```

eprbell (changes requested):
Mostly nits, looks good.

src/dali/plugin/input/csv/bitbank_supplemental.py
macanudo527: This is stripping the optionality.

eprbell: I meant:
```suggestion
        if not self.__withdrawals_code:
            raise Exception("Bitbank.cc supplemental plugin withdrawals file declared without withdrawals code.")
```

After that you should be able to use `self.__withdrawals_code` without optionality.

macanudo527:
This seems good to me.

macanudo527 (commented):
I think this is ready, unless you have something else to add.

eprbell:
Thanks for pinging me on this one: will do a final review in the next few days.

eprbell (approved):
LGTM!

Issue #112, opened 2023-01-24
Issues with USDC->USD pair conversion


kevinkarnani:
Hey all,

Just wanted to show my appreciation for the creation of a tool like this! As a first time user of this tool, I had a few issues while attempting to use it. I have accounts on Coinbase + Coinbase Pro, so I got all the config info down before running into two issues. Initially, while not specifying a pair converter in the `.ini` file, I get the following logs (ran with the DEBUG prefix):

```
2023-01-24 02:07:12,667/CCXT-converter/high/DEBUG: Converting USDC to USD
2023-01-24 02:07:12,834/CCXT-converter/high/DEBUG: Using default exchange Binance.com type for Coinbase.
2023-01-24 02:07:12,986/dali/ERROR: Fatal exception occurred:
Traceback (most recent call last):
  File "~/miniforge3/lib/python3.9/site-packages/ccxt/base/exchange.py", line 622, in fetch
    response.raise_for_status()
  File "~/.local/lib/python3.9/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 451 Client Error:  for url: https://api.binance.com/api/v3/exchangeInfo

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/Desktop/crypto/dali-rp2/src/dali/dali_main.py", line 168, in _dali_main_internal
    resolved_transactions: List[AbstractTransaction] = resolve_transactions(transactions, dali_configuration, args.read_spot_price_from_web)
  File "~/Desktop/crypto/dali-rp2/src/dali/transaction_resolver.py", line 259, in resolve_transactions
    transaction = _update_spot_price_from_web(transaction, global_configuration)
  File "~/Desktop/crypto/dali-rp2/src/dali/transaction_resolver.py", line 136, in _update_spot_price_from_web
    conversion: RateAndPairConverter = _get_pair_conversion_rate(
  File "~/Desktop/crypto/dali-rp2/src/dali/transaction_resolver.py", line 108, in _get_pair_conversion_rate
    rate = cast(AbstractPairConverterPlugin, pair_converter).get_conversion_rate(timestamp, from_asset, to_asset, exchange)
  File "~/Desktop/crypto/dali-rp2/src/dali/abstract_pair_converter_plugin.py", line 129, in get_conversion_rate
[...]

Issue #110, opened 2023-01-22
CCXT Pair Converter may not be available for people in the US


macanudo527:
It seems like the Binance Public API has been geolocked for the US. Since Binance is the default exchange for the CCXT pair converter, pricing might not be available for those using Dali-RP2 in the US.

We need another default exchange. Kraken is good candidate, but fetching pricing is really slow.

eprbell:
What a pain in the neck... Kraken seems like a good alternative and hopefully caching will help with slowness issues.

macanudo527:
It's just a simple line change. But, any prices from the last full quarter on (currently Q42022~) will take 5 seconds to retrieve. 

The KuCoin API seems more responsive and has more markets, but prices start 9/2017. Kraken has a deeper history.

Can you access [Binance's CSV files](https://data.binance.vision/?prefix=data/spot/daily/klines/)? Can you click the link and see if you are locked out? That is another option as well. 

Obviously, it would beneficial for users to set this as a parameter as well.

eprbell:
Should we implement #61 to fix this? That would give users full flexibility to choose what they want, I think. I can access the Binance CSV files you linked.

macanudo527:
This has been resolved with #115 .

Issue #61, opened 2022-07-24
Implement subclasses of CCXT locked to one exchange


macanudo527:
> Actually this is an interesting idea: would it make sense to add variants (subclasses) of your CCXT converter that use a fixed exchange (one for Binance, one for Kraken, etc.)? The binance subclass would be exactly equivalent to the pair converter in this PR. Additionally, if we added a CCXT converter subclass for Coinbase, it should be equivalent to the HistoricCrypto plugin. Does that sound correct or am I missing something?

_Originally posted by @eprbell in https://github.com/eprbell/dali-rp2/issues/60#issuecomment-1193232245_

Implement subclasses or create a method of prioritizing one exchange's pricing data over another.

macanudo527:
Is there a way to simply place two different instantiations of the same plugin in the pricing stack? So that we could just instantiate one CCXT plugin locked to Binance, and then another locked to Coinbase. I don't think we would need to put much in the subclass other than have a different class name.

Or is there an automagic way in Python to create generic subclasses dynamically?

eprbell:
> o

I think the best way is to have separate python files, one per exchange: each file would contain a (probably) trivial subclass of CCXT. Then the user can arrange them in their priority order of choice in the configuration.

macanudo527:
I'm pretty sure this can simply be implemented by passing a param. I don't think a subclass is really needed. Or are subclasses preferred for clarity's sake?

eprbell:
Yeah, I'm thinking classes would be both clearer to use and more in line with the rest of the design (one class per service, whether it's an exchange, a price provider, a wallet, etc.).

macanudo527:
This should be resolved with the latest PR.

PR #115, opened 2023-02-03
Implement subclasses of CCXT locked to one exchange

macanudo527:
The primary goal of this PR is to address #61.

A warning has also been added for those trying to use Binance for pricing in the USA to address #110.
This should also solve #112 by allowing users to lock pricing resolution to one exchange.

Note that the locked classes aren't actually needed. I user can just call the `dali.plugin.pair_converter.ccxt` with the `default_exchange` and `exchange_locked` set to have the same effect.

I'm wondering if there is a more elegant way to handle `call_count` and `call_with` since I needed to disable a lot to get them working. (Lines #491-492 in `test_plugin_ccxt.py`) 

Do I need to declare/type something?

EDIT: Nevermind, I figured out what I was doing wrong. Disregard the comment on `call_count` and `call_with`.

Further documentation is coming.


macanudo527:
Do you want me to handle the too broad exceptions for all the files in this PR? I guess I need to update pylint locally.

I'm guessing I need to open another one. Should I replace `Exception` with `RP2Exception`?

src/dali/plugin/pair_converter/ccxt_binance_locked.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2023 Neal Chambers
```
eprbell, line 1: Just a nit: why not call this just ccxt_binance.py: is there ever going to be a ccxt_binance.py that is not locked?

src/dali/plugin/pair_converter/ccxt_kraken_locked.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2023 Neal Chambers
```
eprbell, line 1: Just a nit: why not call this just ccxt_kraken.py: is there ever going to be a ccxt_kraken.py that is not locked?

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -196,73 +216,90 @@ def _bfs_cyclic(self, graph: Dict[str, Dict[str, None]], start: str, end: str) -
         # No path found
         return None
 
-    def get_historic_bar_from_native_source(self, timestamp: datetime, from_asset: str, to_asset: str, exchange: str) -> Optional[HistoricalBar]:
```
eprbell, line 199: From here on the diff becomes a bit hard to read, probably because some code got moved around, causing the diff to become more complex. Is there any chance you could restore the original order of the code and then do another PR with code reorg later?

eprbell (changes requested):
Great to see locked exchanges coming along! Here's some initial feedback.

eprbell:
> Do you want me to handle the too broad exceptions for all the files in this PR? I guess I need to update pylint locally.
> 
> I'm guessing I need to open another PR. Should I replace `Exception` with `RP2Exception`?

Good point: I think the correct solution would be to add RP2RuntimeError in RP2 at https://github.com/eprbell/rp2/blob/main/src/rp2/rp2_error.py, and then substitute all occurrences of Exception in the code with RP2RuntimeError.

src/dali/plugin/pair_converter/ccxt.py
macanudo527: I needed to bundle this up into a function in order to test it. Maybe if I move it to the end?

eprbell: I think if you keep the same function order as in the original file, the diff might become easier to read. Let's try that: if it doesn't work I'll manage.

macanudo527: I think what happened is the new function `_add_exchange_to_memcache` looks too similar to `get_historic_bar_from_native_source` and they are in the same place. I moved the function toward the end. I'm not sure if that helps or if I just made a bigger mess of it.

I did start using Sublime Merge to help me see if the diff is too out of wack before I commit. I'll try to keep an eye on it.

macanudo527:
> > Do you want me to handle the too broad exceptions for all the files in this PR? I guess I need to update pylint locally.
> > I'm guessing I need to open another PR. Should I replace `Exception` with `RP2Exception`?
> 
> Good point: I think the correct solution would be to add RP2RuntimeError in RP2 at https://github.com/eprbell/rp2/blob/main/src/rp2/rp2_error.py, and then substitute all occurrences of Exception in the code with RP2RuntimeError.

So, open another PR and merge that first?

macanudo527:
I submitted eprbell/rp2#89 to address the too broad exception errors. Once that gets merged, I'll submit another for Dali.

eprbell:
> I submitted [eprbell/rp2#89](https://github.com/eprbell/rp2/pull/89) to address the too broad exception errors. Once that gets merged, I'll submit another for Dali.

It's merged: use RP2 1.4.1 in the DaLI setup.cfg file and you should be able to see the new exception in DaLI. Let me know if there are any problems.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -460,3 +434,52 @@ def find_historical_bar(self, from_asset: str, to_asset: str, timestamp: datetim
             self._add_bar_to_cache(key, result)
 
         return result
+
+    def _add_alternative_markets(self, current_graph: Dict[str, Dict[str, None]], current_markets: Dict[str, List[str]]) -> None:
+        for base_asset, quote_asset in _ALTMARKET_BY_BASE_DICT.items():
```
eprbell, line 439: _ALT_MARKET_BY_BASE_DICT

eprbell (changes requested):
Couple of nits, otherwise it looks good.

macanudo527:
Just need to merge #117 and this should be good to go.

eprbell:
> Just need to merge #117 and this should be good to go.

Sounds good. I merged #117: could you bring that in this branch and run tests again? Then I'll finish the review and we can merge this too.

eprbell:
> > Just need to merge #117 and this should be good to go.
> 
> Sounds good. I merged #117: could you bring that in this branch and run tests again? Then I'll finish the review and we can merge this too.

Oops, I thought I merged it but somehow the PR was still open. Merged it now.

macanudo527:
This passed locally. Let's see if github likes it.

eprbell (approved):
LGTM. Thanks for all this good work!

PR #118, opened 2023-02-09
Add Optional to exchange_locked

macanudo527:
Sorry about that. I forgot about the transaction resolver checking for `Optional`. This should fix the new concern with #112.


eprbell (approved):
LGTM

PR #119, opened 2023-02-11
Improve File Handling in Kraken CSV

macanudo527:
This fixes some alternative market handling in the CCXT pair converter as well as allow the pair converter to continue with REST price resolution if the file isn't found on Google Drive like what is described in #112 

EDIT: Let me do a full pull and double-check if all alternative markets are added. That could be ~12 hours.

### DIFF
--- a/src/dali/plugin/pair_converter/ccxt.py
+++ b/src/dali/plugin/pair_converter/ccxt.py
@@ -83,8 +83,10 @@ _ALT_MARKET_EXCHANGES_DICT: Dict[str, str] = {
     "ATDUSDT": _GATE,
     "BSVUSDT": _GATE,
     "BOBAUSD": _GATE,
+    "BUSDUSDT": _BINANCE,
     "EDGUSDT": _GATE,
     "ETHWUSD": _KRAKEN,
+    "NEXOUSDT": _BINANCE,
     "SGBUSD": _KRAKEN,
     "SOLOUSDT": _HUOBI,
     "USDTUSD": _KRAKEN,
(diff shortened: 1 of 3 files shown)


macanudo527:
I spoke too soon. I was only missing two other markets. We will most likely have to work out a better solution to just manually adding these though.

eprbell:
Sounds good, thanks for jumping on this!

macanudo527:
This still results in the program self-terminating when pulling pricing for 6+ different crypto assets on my machine with 40gb a RAM. I will open another PR soon to better handle CSV caching.

macanudo527:
Do you want to merge this PR for now, and then I can open another that splits up the CSVs since that will be a pretty big revision?

eprbell:
> Do you want to merge this PR for now, and then I can open another that splits up the CSVs since that will be a pretty big revision?

Ok, will review in the next couple of days.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -240,8 +248,8 @@ def get_historic_bar_from_native_source(self, timestamp: datetime, from_asset: s
                 self.__logger.debug("Using default exchange %s type for %s.", self.__default_exchange, exchange)
             exchange = self.__default_exchange
 
-        # Caching of exchanges
-        if exchange not in self.__exchanges:
+        # The exchange could have been added as an alt if so markets wouldn't have been built
```
eprbell, line 251: as an alt: if so

eprbell (changes requested):
LGTM: just a nit in a comment.

macanudo527:
This should be okay now. I have the chunking PR ready, just want to run it in the real-world and see if it holds up.

eprbell (approved):
LGTM

PR #123, opened 2023-03-06
Save and chunk CSV files locally to avoid running out of memory

### LINKED ISSUE #120: Kraken CSV reader creates too large of cache for multiple assets
If a user has more than about 3 assets, pickling the data files uses a lot of memory 30gb+ and may cause the program to terminate.

I think I'll have to work out some kind of cache for just the kraken CSV.

Another option is to break the CSV files down into chunks of time and load only the chunk we need. These files would be saved locally. I don't think this poses a security risk. Does anyone know if it does or see an issue with saving the files locally?

macanudo527:
This resolves #120 

I disabled the rmdir in the `test_ods_output_diff.py` It was too destructive. The test passes without it. However, a future test could of course interfere. Is there a way to be more selective about it? Every time tests are run it flushes cache and long 3+ hour pulls have to be restarted. This is not ideal.


eprbell:
Removing the cache is important for tests because it makes them deterministic: if you allow the cache to be present you can never know what's in the cache, so you can't know which execution paths will be taken and the tests become non-deterministic. 

The 3 hour pull is related to your actual taxes, not to tests, correct? If so I would suggest using two installations of DaLI in two different directories: one for development and the other for normal use. This way your normal-use cache won't get affected. And you can always copy the cache from the normal-use directory to the development one at will, if you wish to do so. Hope this helps.

macanudo527:
Ok that makes sense. I do run tests on this codebase to see how it works with real-world data, but maybe I can copy over cache if I need it. Or just remember to back it up before I run tests.

eprbell:
> Ok that makes sense. I do run tests on this codebase to see how it works with real-world data, but maybe I can copy over cache if I need it. Or just remember to back it up before I run tests.

Sounds good: I'll review this in the next few days.

tests/test_plugin_ccxt.py
```diff
@@ -396,12 +396,16 @@ def test_fiat_pair(self, mocker: Any) -> None:
...
+        cache_path = os.path.join(CACHE_DIR, "Test-" + plugin.cache_key())
         if os.path.exists(cache_path):
             os.remove(cache_path)
 
         kraken_csv = KrakenCsvPricing(google_api_key="whatever")
-        with open("input/test_kraken_CSV.zip", "rb") as file:
+        mocker.patch.object(kraken_csv, "cache_key").return_value = "Test-" + kraken_csv.cache_key()
+        mocker.patch.object(kraken_csv, "_Kraken__CACHE_DIRECTORY", "input/kraken_test")
+        if not os.path.exists("input/kraken_test"):
+            os.makedirs("input/kraken_test")
```
eprbell, line 407: Does this write in the top level input directory? Any test-generated files and directories should go into the output directory.

tests/test_plugin_kraken_csv_download.py
```diff
@@ -0,0 +1,67 @@
...
+                pass
+
+        cache_path = path.join(CACHE_DIR, kraken_csv.cache_key())
+        if path.exists(cache_path):
+            remove(cache_path)
+
+        mocker.patch.object(kraken_csv, "_Kraken__CACHE_DIRECTORY", _CACHE_DIRECTORY)
+        if not path.exists(_CACHE_DIRECTORY):
+            makedirs(_CACHE_DIRECTORY)
+        with open("input/USD_OHLCVT_test.zip", "rb") as file:
```
eprbell, line 51: Just wanted to make sure that input/USD_OHLCVT_test.zip has made-up data in it and not pasted data from their files: if we're going to ship it with our distro it has to contain fake data we use for testing.

eprbell (commented):
Looks good. I did a quick initial review to give you some feedback, but I'll do a more detailed one in the weekend.

tests/test_plugin_kraken_csv_download.py
macanudo527: oh curses, yeah, stupid licensing. I'll edit the amounts to be completely made up.

tests/test_plugin_ccxt.py
macanudo527: Yeah, output-ed files should go in output. Go figure. Thanks. :)

macanudo527:
3rd time a charm I guess. Did you have anything else to check on this?

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -59,13 +62,40 @@
...
+_TIME_GRANULARITY: List[str] = [_MINUTE, _FIVE_MINUTE, _FIFTEEN_MINUTE, _ONE_HOUR, _TWELVE_HOUR, _ONE_DAY]
+
+# Chunking variables
+_PAIR_START: str = "start"
+_PAIR_MIDDLE: str = "middle"
+_PAIR_END: str = "end"
+_MAX_MULTIPLIER: int = 500
+
+
+class PairStartEnd(NamedTuple):
```
eprbell, line 87: This looks like a private class: if so, prepend with `_`

eprbell (changes requested):
Just one more nit found. Otherwise LGTM.

src/dali/plugin/pair_converter/csv/kraken.py
macanudo527: You meant protected right?

eprbell: Protected can only be attributed to members of a class. In this case we're referring to a class name which is in the global scope, so it can only be public or private wrt to the scope it's in.

eprbell (approved):
LGTM!

Issue #125, opened 2023-03-10
Does dali-rp2 have a .ods plugin for previous outputs?


jamesbaber1:
Hi, thanks for all the work on this project. This is awesome.

Maybe I missed this in the docs, but my question is this, is there a way to load in data from your existing .ods output as an input?

Here is my example. I used dali to collect all my trades and price data for 2021. Due to some various complications I had to manually resolve some transactions in the .ods output file before then feeding that into rp2. So I saved the corrected 2021 .ods file.

Now I collected my new inputs using dali, using some of the awesome rest and csv plugins for 2022.  But I would also like to feed in my original 2021 .ods file and combine with the new inputs. 

Is there a way to do this already? Or is there an easy way to combine two .ods files?

Thanks!

eprbell:
Thanks! Take a look at the [manual plugin](https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md#manual-section-csv). It can be used for multiple things:
* resolve partial transactions (without needing to modify the output .ods), or
* read back DaLI output (after some small modifications)

I recommend the first approach, because it doesn't require modifying the DaLI .ods output.

Hope this helps!

eprbell:
No modifications to DaLI source. What I mean is that you can take DaLI output, split it into 3 .csv files (in, out, intra), move the unique_id column to position 0 and feed the result to the manual plugin.

jamesbaber1:
Ahh ok gotcha. Its seems relatively straight forward to write a input plugin after looking at 
https://github.com/eprbell/dali-rp2/blob/424a7e92071244df5434b526e6c074dda548c432/src/dali/plugin/input/csv/manual.py#L37

Would there by any cocerns will me opening a PR with an input ods plugin. (that supports only the rp2 input .ods format)?

I have been meaning to contribute to this at some point, and it seems like an easy way to get more familiar with it

eprbell:
That's an interesting idea: it may be an alternative to the manual plugin, for people who prefer it. Sounds good to me. You may want to look at the [RP2 ods parser](https://github.com/eprbell/rp2/blob/main/src/rp2/ods_parser.py): the code will be very similar (maybe after you're done, we can think about how to factor it).

jamesbaber1:
Ok cool, yea I am actually importing most of all the logic from rp2 and the ods parser. Its looking like it wont be much code at all.

[...]

PR #126, opened 2023-03-11
Existing RP2 Input (ODS) Plugin

jamesbaber1:
Created this PR after discussing the plugin here
https://github.com/eprbell/dali-rp2/issues/125

## How it works
If you have the existing RP2 input files (The config and data in the .ods file generated by dali-rp2). You can use that as input for dali if you would like to 
generate a new output that is combined with other plugins you pair with it. An example might be; you want to combine your old data used from a previous
tax return with new data, or you want to combine multiple rp2 configs and data inputs into a single .ods file and single config file.

### configuration
Here is an example of how to configure the plugin
```ini
[dali.plugin.input.ods.rp2_input]
configuration_path = input/bob_2021_input.ini
input_file = input/bob_2021_input.ods
country=us
from_date=2021-01-01T00:00:00 ;optional
to_date=2021-12-31T00:00:00 ;optional
native_fiat=USD ;optional
```


src/dali/plugin/input/ods/rp2_input.py
```diff
@@ -0,0 +1,164 @@
...
+
+            input_data: InputData = parse_ods(
+                configuration=configuration,
+                asset=asset,
+                input_file_handle=input_file_handle
+            )
+            self.__logger.debug("InputData object: %s", input_data)
+            for asset_entry in input_data.unfiltered_in_transaction_set:
+                transaction: RP2InTransaction = cast(RP2InTransaction, asset_entry)
+                result.append(InTransaction(
```
eprbell, line 107: Add raw data log (useful for debugging):
```
self.__logger.debug("Transaction: %s", str(transaction))
```

src/dali/plugin/input/ods/rp2_input.py
```diff
@@ -0,0 +1,164 @@
...
+                    crypto_fee=str(transaction.crypto_fee) if transaction.crypto_fee else None,
+                    fiat_in_no_fee=str(transaction.fiat_in_no_fee) if transaction.fiat_in_no_fee else None,
+                    fiat_in_with_fee=str(transaction.fiat_in_with_fee) if transaction.fiat_in_with_fee else None,
+                    fiat_fee=str(transaction.fiat_fee) if transaction.fiat_fee else None,
+                    notes=str(transaction.notes) if transaction.notes else None
+                ))
+
+            for asset_transfer in input_data.unfiltered_intra_transaction_set:
+                transaction: RP2IntraTransaction = cast(RP2IntraTransaction, asset_transfer)
+                result.append(
```
eprbell, line 127: Add raw data log (useful for debugging):
```
self.__logger.debug("Transaction: %s", str(transaction))
```

src/dali/plugin/input/ods/rp2_input.py
```diff
@@ -0,0 +1,164 @@
...
+                        to_holder=transaction.to_holder,
+                        spot_price=str(transaction.spot_price) if transaction.spot_price else None,
+                        crypto_sent=str(transaction.crypto_sent),
+                        crypto_received=str(transaction.crypto_received),
+                        notes=str(transaction.notes) if transaction.notes else None
+                    ))
+
+            for asset_exit in input_data.unfiltered_out_transaction_set:
+                transaction: RP2OutTransaction = cast(RP2OutTransaction, asset_exit)
+                result.append(
```
eprbell, line 146: Add raw data log (useful for debugging):
```
self.__logger.debug("Transaction: %s", str(transaction))
```

tests/test_plugin_ods_rp2_input.py
```diff
@@ -0,0 +1,50 @@
...
+from dali.in_transaction import InTransaction
+from dali.out_transaction import OutTransaction
+from dali.plugin.input.ods.rp2_input import InputPlugin
+
+
+class TestRP2InputOds:
+    def test_btc_transactions(self) -> None:
+        plugin = InputPlugin(
+            configuration_path='input/test_ods_rp2_input.ini',
+            input_file='input/test_ods_rp2_input.ods',
```
eprbell, line 27: Can you use a file with two (or more) assets (e.g. [this one](https://github.com/eprbell/rp2/blob/main/input/crypto_example.ods) or similar)?

tests/test_plugin_ods_rp2_input.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        result = plugin.load()
+        assert len(result) == 6
+
+        btc_transaction_in_1: InTransaction = result[0]  # type: ignore
+        # btc_transaction_in_2: InTransaction = result[1]  # type: ignore
+        # btc_transaction_intra_1: InTransaction = result[2]  # type: ignore
+        # btc_transaction_intra_2: InTransaction = result[3]  # type: ignore
+        # btc_transaction_intra_3: InTransaction = result[4]  # type: ignore
+        # btc_transaction_out_1: OutTransaction = result[5]  # type: ignore
```
eprbell, line 39: Can you remove commented code (either delete or uncomment, but it would be good to test a few more transaction)?

eprbell (changes requested):
This is good code! A lot simpler than I even thought initially. I have some minor comments here and there and also take a look at the static analysis warnings.

src/dali/plugin/input/ods/rp2_input.py
```diff
@@ -0,0 +1,164 @@
...
+            elif country.lower() == 'jp':
+                self.__country = jp.JP()
+            else:
+                ValueError(f'The country code {country} is not a valid country code supported by RP2')
+        # otherwise default to us
+        else:
+            self.__country = us.US()
+
+        # validate the country
+        AbstractCountry.type_check("country", self.__country)
```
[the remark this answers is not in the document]
jamesbaber1: Added COUNTRY to the default config. Not sure if is what you were meaning, but take a look and let me know.

tests/test_plugin_ods_rp2_input.py
jamesbaber1: Ok I will add the ETH transactions to this

src/dali/plugin/input/ods/rp2_input.py
```diff
@@ -0,0 +1,137 @@
...
+        input_file: str,
+        native_fiat: Optional[str] = None,
+    ) -> None:
+        super().__init__(account_holder="", native_fiat=native_fiat)
+        self.__configuration_path: str = configuration_path
+        self.__input_file: str = input_file
+        self.__logger: logging.Logger = create_logger(self.__RP2_INPUT)
+        self.__country: AbstractCountry = DEFAULT_CONFIGURATION[Keyword.COUNTRY.value]
+        self.__from_date: date = MIN_DATE
+        self.__to_date: date = MAX_DATE
```
eprbell, line 51: Nit: same for `self.__from_date` and `self.__to_date`: you could just use `MIN_DATE` and `MAX_DATE` directly.

eprbell (changes requested):
Almost there: just a few more changes and then we can merge. Also check the new run of static analysis.

src/dali/dali_main.py
```diff
@@ -63,6 +63,7 @@ class _InputPluginHelperArgs(NamedTuple):
 
 
 def dali_main(country: AbstractCountry) -> None:
+    DEFAULT_CONFIGURATION[Keyword.COUNTRY.value] = country
```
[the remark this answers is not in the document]
jamesbaber1: Ok I updated. I am still not sure I am understanding what your meaning by "global". I didn't think you meant just stuffing the python globals, cause I didn't seem to see this being done anywhere else in the code. So I delcared COUNTRY in the scope of the configuration module, and then `configuration.COUNTRY` is patched directly in `dali_main()` so it can be accessed globally but is still namespaced under the dali configuration.

I am probably misunderstanding what you are wanting

src/dali/configuration.py
```diff
@@ -207,6 +207,8 @@ def type_check_from_string(cls, keyword: str) -> "Keyword":
     },
 }
 
+COUNTRY: AbstractCountry = None
```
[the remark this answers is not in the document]
eprbell: This is good (it's what I meant). Actually I thought of another way that would be better and doesn't require the dependency to configuration: just add country as a parameter to the load method and pass it in (this requires changing the load methods of all plugins). However this is good enough for now, if you don't have time to change `load()` (we can change it separately in a separate upload).

eprbell (changes requested):
Looks pretty good. Be sure to also incorporate the results of static analysis.

src/dali/configuration.py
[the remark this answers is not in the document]
jamesbaber1: Ok cool that sounds good to me. Ill fix up the last changes and let you decide what would be the best way to pass in the country and you can update in another PR.

eprbell (approved):
LGTM: thanks for working on this!

eprbell:
I merged the [changes discussed above](https://github.com/eprbell/dali-rp2/commit/ec9d4e9c35ba4a88a7310e9372c918e4998f1d84): the `load()` method now receives the country as an argument.

PR #138, opened 2023-04-05
Add visual studio project

kiates:
It took me a bit of fiddling, since I don't know what I'm doing when it comes to python, but I was able to get a visual studio project going that supports running the UTs, and starting via the debugger.

I did have to make some changes since I don't know of any way to work-around the console scripts that are installed for the country plugins to bootstrap things.

I'm not sure if there is another way, but at least with this I seem to be able to have access to tools I'm more used to. This will help me to better understand flow and maybe be in a better position to contribute going forward. Maybe this will be useful to others?


macanudo527:
Could you take the Pionex Documentation commit off of this?

dali-rp2.pyproj
```diff
@@ -0,0 +1,135 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <Content Include="config\test_config.ini" />
+    <Content Include="input\test_ods_rp2_input.ini" />
+    <Content Include="mypy.ini" />
+    <Content Include="src\dali_rp2.egg-info\dependency_links.txt" />
+    <Content Include="src\dali_rp2.egg-info\entry_points.txt" />
+    <Content Include="src\dali_rp2.egg-info\requires.txt" />
+    <Content Include="src\dali_rp2.egg-info\SOURCES.txt" />
+    <Content Include="src\dali_rp2.egg-info\top_level.txt" />
```
eprbell, line 35: These files are generated: I'm not sure they should be included in an IDE config file.

dali-rp2.pyproj
```diff
@@ -0,0 +1,135 @@
...
+    <ProjectGuid>{21de6965-e7a9-45bf-bc35-9aafe37f7ddf}</ProjectGuid>
+    <ProjectHome />
+    <StartupFile>src\dali\plugin\country\us.py</StartupFile>
+    <SearchPath />
+    <WorkingDirectory>.</WorkingDirectory>
+    <OutputPath>.</OutputPath>
+    <ProjectTypeGuids>{888888a0-9f3d-457c-b088-3a5042f75d52}</ProjectTypeGuids>
+    <LaunchProvider>Standard Python launcher</LaunchProvider>
+    <InterpreterId>MSBuild|.venv|$(MSBuildProjectFullPath)</InterpreterId>
+    <CommandLineArguments>-t 1 -s -o C:\Users\cyates\Documents\rp2-files\cyates\dali-output -p cyates- C:\Users\cyates\Documents\rp2-files\cyates\cyates.ini</CommandLineArguments>
```
eprbell, line 15: This line has paths that point to a specific directory on your machine: it should be generic instead.

dali-rp2.pyproj
```diff
@@ -0,0 +1,135 @@
...
+    <Folder Include="src\dali\data" />
+    <Folder Include="src\dali\plugin" />
+    <Folder Include="src\dali\plugin\country" />
+    <Folder Include="src\dali\plugin\input" />
+    <Folder Include="src\dali\plugin\input\csv" />
+    <Folder Include="src\dali\plugin\input\ods" />
+    <Folder Include="src\dali\plugin\input\rest" />
+    <Folder Include="src\dali\plugin\pair_converter" />
+    <Folder Include="src\dali\plugin\pair_converter\csv" />
+    <Folder Include="src\dali_rp2.egg-info" />
```
eprbell, line 120: Also generated.

src/dali/plugin/country/jp.py
```diff
@@ -21,3 +21,8 @@
 # JP-specific entry point
 def dali_entry() -> None:
     dali_main(JP())
+
```
eprbell, line 24: IDE integration shouldn't require modifying the source code. I usually debug with print, but I did use the VSCode debugger occasionally in the past (not sure how similar it is to Visual Studio). VSCode can be configured to debug DaLI and RP2 without changing the source: I used .venv/bin/dali_us as the entry point.

src/dali/plugin/country/us.py
```diff
@@ -21,3 +21,8 @@
 # US-specific entry point
 def dali_entry() -> None:
     dali_main(US())
+
```
eprbell, line 24: Same as above.

dali-rp2.pyproj
```diff
@@ -0,0 +1,135 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectGuid>{21de6965-e7a9-45bf-bc35-9aafe37f7ddf}</ProjectGuid>
```
eprbell, line 6: What is this id? Can it be shared on other user's machines?

dali-rp2.pyproj
```diff
@@ -0,0 +1,135 @@
...
+    <ProjectHome />
+    <StartupFile>src\dali\plugin\country\us.py</StartupFile>
+    <SearchPath />
+    <WorkingDirectory>.</WorkingDirectory>
+    <OutputPath>.</OutputPath>
+    <ProjectTypeGuids>{888888a0-9f3d-457c-b088-3a5042f75d52}</ProjectTypeGuids>
```
eprbell, line 12: What is this id? Can it be shared on other user's machines?

eprbell (changes requested):
I like the idea of providing IDE-specific configuration files, but this may require a bit more work (see comments).

Also, I'd like to move all IDE specific files to config/ide/<ide>/, so that they don't crowd the top directory, but they are easy to access.

dali-rp2.pyproj
kiates: When I created the project I used Microsoft's "import" existing python project.  That's probably why it included these. I've taken them out.

dali-rp2.pyproj
kiates: I pointed this to the sample project.

src/dali/plugin/country/jp.py
kiates: I can create IDE specific dali_us/dali_jp.py files for debugging, but I don't think that will be as simple as just adding this entry point.  I couldn't figure out any way to get visual studio to start using the exe's generated by the "installer".

dali-rp2.pyproj
kiates: This is a unique identify that represents this pyproject if it is inserted into a solution file.  It isn't user specific.

dali-rp2.pyproj
kiates: This is a unique ID defined by Microsoft that indicates that this project is a python project type. It is the same for all python projects.

kiates:
> 

I tried some solutions to what your requesting, and I think it is more trouble than it is worth to fight how Visual Studio expects a python project to look like.  In order to move the project/solution (pythonproject/sln) files to a sibling directory it requires changing all reference to source to a "linked" file rather than a normal file.  This very robust, requires hand editing the files, and is a bit of a nightmare to maintain while breaking some of the IDE functionality to include/exclude source files.  In addition the tests would no longer run for me (there may be a solution for that, but given the other problems I don't recommend trying to fight VS).

Also, if it was easy to have it in a sibling directory I would also recommend putting it in a ide or visual-studio/visual_studio directory rather than config as in config it is kind of hidden away in a directory that isn't related.

If you want to see my "experimentation" trying to get this going how you envision, see this branch in my fork: https://github.com/kiates/dali-rp2/tree/visual-studio-project-improvements.

If you aren't willing to take this in, I can just reserve on my fork for when I'm working with the project. It would just be nice to not have to maintain a fork, and could be useful to other VS contributors.

src/dali/plugin/country/jp.py
eprbell: Have you tried using .venv/bin/dali_us and .venv/bin/dali_jp as entry points? They should not require modification.

eprbell:
> > 
> 
> I tried some solutions to what your requesting, and I think it is more trouble than it is worth to fight how Visual Studio expects a python project to look like. In order to move the project/solution (pythonproject/sln) files to a sibling directory it requires changing all reference to source to a "linked" file rather than a normal file. This very robust, requires hand editing the files, and is a bit of a nightmare to maintain while breaking some of the IDE functionality to include/exclude source files. In addition the tests would no longer run for me (there may be a solution for that, but given the other problems I don't recommend trying to fight VS).
> 
> Also, if it was easy to have it in a sibling directory I would also recommend putting it in a ide or visual-studio/visual_studio directory rather than config as in config it is kind of hidden away in a directory that isn't related.
> 
> If you want to see my "experimentation" trying to get this going how you envision, see this branch in my fork: https://github.com/kiates/dali-rp2/tree/visual-studio-project-improvements.
> 
> If you aren't willing to take this in, I can just reserve on my fork for when I'm working with the project. It would just be nice to not have to maintain a fork, and could be useful to other VS contributors.

I think it's worth experimenting in the direction you're going, but I would like to avoid IDE files in the root dir and IDE-specific changes to the source. The reason is that if we do that for every IDE the source tree will quickly become messy. 

So I see the following options:
* keep experimenting with it and we can revisit later, if you have a new version that meets the above requirements;
* document it instead: we can have a file in docs describing what to do for various IDEs;
* fork it (which is what you have done already).

PR #139, opened 2023-04-05
Documentation updates

kiates:
This looked like a copy/paste error.

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -597,7 +597,7 @@ fiat_fee = 11
 unique_id = 12
 notes = 13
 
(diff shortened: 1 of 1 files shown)


kiates:
Did the linking system change for ccxt today?  I tried to follow the link (https://docs.ccxt.com/en/latest/manual.html#instantiation) and got a 404.  This link works though: https://docs.ccxt.com/#/?id=instantiation.  There are other links that seems to have the same issue, but they weren't reported (maybe the checker aborts on the first issue?)

eprbell (approved):
Thanks for fixing!

PR #142, opened 2023-04-08
Implement Djikstra in CCXT Pair Converter

macanudo527:
This is the implementation of the Djikstra algo in the CCXT Pair Converter.

I'd like some feedback on the weights. I think these are fair, but if you could imagine a scenario where these don't work let me know.

I'm also wondering about vertexes. I pop them out of `Graph` with a list comprehension in each function, but I'm guessing this should be a function somewhere. Should it be a class method in `ccxt.py` or should we push it down into prezzemolo? I'm guessing class method.

**This is WIP.** 
I should run it with real-world data, which will take some time, but if all goes well, it should finish over the weekend.


src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -134,6 +136,11 @@
 # CSV Reader
 _GOOGLE_API_KEY: str = "google_api_key"
 
+# Djikstra weights
+# Priority should go to quote assets listed above, then other assets, and finally alternatives
+_STANDARD_WEIGHT: float = 50.0
+_ALTERNATIVE_MARKET_WEIGHT: float = 50.5
```
eprbell, line 142: Just a nit: I would avoid decimals, if possible, just for esthetics.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -179,35 +198,32 @@ def _build_fiat_list(self) -> None:
             LOGGER.info("Fetching of fiat symbols failed. The server might be down. Please try again later.")
             raise RP2RuntimeError("JSON decode error") from exc
 
-    def _add_fiat_edges_to_graph(self, graph: Dict[str, Dict[str, None]], markets: Dict[str, List[str]]) -> None:
+    def _add_fiat_edges_to_graph(self, graph: Graph[str], markets: Dict[str, List[str]]) -> None:
+        vertexes: Dict[str, Vertex[str]] = {v.name: v for v in graph.vertexes}
```
eprbell, line 202: Regarding your question on vertexes, I would avoid adding the mapping to the Graph class in prezzemolo because we're trying to keep it small: this mapping can be easily managed in the application (which is adding the vertexes after all). So I would have a class dictionary here that manages name -> vertex (and some protected accessors to make it usable in subclasses). Basically every time you add a new vertex to the graph you would also update the dictionary.

eprbell (commented):
Exciting to see this! I gave it a very quick review but I will go deeper after you run with real-world data.

src/dali/plugin/pair_converter/ccxt.py
macanudo527: Are the weights `float` in `prezzemolo` to facilitate calculations (as opposed to set values like these) then?

eprbell: Nothing wrong with using decimals: it was just an esthetic nit, not a strong opinion. Feel free to skip it if you prefer.

macanudo527:
I'm not sure how to proceed. I'm not getting this error locally.

I have python 3.10.6 in my virtualenv, but the CI is running 3.8.16. I think that is what is causing it. Is there a way to safely switch python versions, or a way to make mypy happy at the lower version?

macanudo527:
Maybe I'm just being too fancy for my own good. The whole point was to add to `Graph` whenever the `defaultdict` found a missing `Vertex` all in one function. I was trying to avoid conditionals. I'll just have to use them I guess.

eprbell:
> Maybe I'm just being too fancy for my own good. The whole point was to add to `Graph` whenever the `defaultdict` found a missing `Vertex` all in one function. I was trying to avoid conditionals. I'll just have to use them I guess.

You can still do the same with the following approach. Add the following to `AbstractPairConverterPlugin`:
* `__vertexes`: the dictionary
* `_get_vertex(name)`
* `_add_vertex(name, vertex)`
* `_get_or_set_vertex(name, vertex)`: equivalent to `setdefault`
* whatever else you may need...

What conditionals are you referring to?

macanudo527:
> > Maybe I'm just being too fancy for my own good. The whole point was to add to `Graph` whenever the `defaultdict` found a missing `Vertex` all in one function. I was trying to avoid conditionals. I'll just have to use them I guess.
> 
> You can still do the same with the following approach. Add the following to `AbstractPairConverterPlugin`:
> 
> * `__vertexes`: the dictionary
> * `_get_vertex(name)`
> * `_add_vertex(name, vertex)`
> * `_get_or_set_vertex(name, vertex)`: equivalent to `setdefault`
> * whatever else you may need...
> 
> What conditionals are you referring to?

Basically, I'll have to recode defaultdict. Psuedo-code:

self.__vertexes.get(current_vertex) # This is already built into defaultdict
if not current_vertex:                     # This line, too
    current_vertex = Vertex[str](name="name")
    current_graph.add_vertex(current_vertex)
   self.__vertexes["name"] = current_vertex

I was trying to overload defaultdict since it is already in there, and it works. Just Python 3.7.16 doesn't seem to recognize mypy type declarations in class inheritance. 

These methods will have to be added to a data class. If they are added to `AbstractPairConverterPlugin` the graph would have to be passed to each method. Currently, each exchange has its own graph. I'd like to implement one universal graph, but that will take a little more architecture. Like I said, I was trying to be too fancy.

tests/test_abstract_pair_converter.py
```diff
@@ -0,0 +1,37 @@
...
+
+from dali.abstract_pair_converter_plugin import GraphVertexesDict
+
+
+class TestAbstractPairConverterPlugin:
+    def test_graph_vertexes_class(self) -> None:
+        current_graph: Graph[str] = Graph[str]()
+        graph_vertexes: DefaultDict[str, Vertex[str]] = GraphVertexesDict(current_graph)
+
+        first_vertex = graph_vertexes["first vertex"]
```
eprbell, line 28: This approach is clever but I'm afraid it makes the code less readable, because you redefined a common behavior of dictionaries: if `graph_vertexes` was a plain dict, then `graph_vertexes["first vertex"]` would have no side effects. But the `GraphVertexesDict` adds side effects under the hood that would change the default semantics of dict in a way makes it more difficult to follow the logic for somebody new to the code (and I argue, even for somebody who's familiar with the semantics of `GraphVertexesDict`, but who hasn't worked on the code for a while). This would throw off most Python programmers: it is a bit like operator redefinition, which is fine in certain limited circumstances, but it can lead to trouble understanding and maintaining the code down the road. 

I would prefer an approach that makes the operations explicit and clear to the reader, without redefining standard behavior. Here's a potential approach based on inheritance, that doesn't modify standard semantics:
```
class MappedGraph (Graph(Generic[ValueType])):
    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None) -> None:
        super().__init__(vertexes)
        self.__name_to_vertex: Dict[str, Vertex[ValueType]] = {vertex.name:vertex for vertex in vertexes}

    def get_vertex(self, name:str) -> Vertex[ValueType]:
        return self.__name_to_vertex[name]

    def add_neighbor(self, vertex: "Vertex[ValueType]", weight: float = 0.0) -> None:
        super().add_neighbor(vertex, weight)
        self.__name_to_vertex[vertex.name] = vertex
```

Another approach would be to use composition, but probably the inheritance-based one is more compact.

eprbell (commented):
Not a full review: I just wanted to give you quick feedback on the new Graph class.

tests/test_abstract_pair_converter.py
eprbell: > And then add a get_or_set method for creating new vertexes:
> 
> ```
> def get_or_set_vertex(self, name:str) -> Vertex[ValueType]:
>       if self.get_vertex(name):
>             return self.get_vertex(name)
>       else:
>             new_vertex: Vertex[ValueType] = Vertex[str](name=name)
>             super().add_vertex(new_vertex)
>             self.__name_to_vertex[name] = new_vertex
>             return new_vertex
> ```

Sounds good.

> And shouldn't `get_vertex` be:
> 
> ```
> def get_vertex(self, name:str) -> Optional[Vertex[ValueType]]:
>         return self.__name_to_vertex.get(name)
> ```
> 
> So that it mimics `get()` behavior?

Looks good, too.

> And, the `add_neighbor()` would be:
> 
> ```
>     def add_neighbor(self, vertex_name: str, neighbor_name: str, weight: float = 0.0) -> None:
>         vertex: Vertex[str] = self.get_or_set_vertex(vertex_name)
>         neighbor: Vertex[str] = self.get_or_set_vertex(neighbor_name)
>         if neighbor not in vertex.neighbors:
>             vertex.add_neighbor(neighbor, weight)
> ```

Looks great, but be change the neighbor check to:
```
         if not vertex.has_neighbor(neighbor):
             vertex.add_neighbor(neighbor, weight)
```
Because it's hashtable-based search rather iterator-based (so it's much faster).

> This is excellent since it eliminates the need for two `dict`s. We just need one that tracks graphs and vertexes together.

Awesome!

macanudo527: This will most likely grow in size since I might as well add `markets` (ie. edges) to this as well. I'd like to implement a priorityqueue that prioritizes markets from exchanges with higher volume. However, that is out of scope for this PR. 

Should I separate it out now into its own file?

eprbell: Do you mean size as is lines of code? You're thinking of separating MappedGraph into a separate class?

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -53,6 +67,31 @@ class AssetPairAndTimestamp(NamedTuple):
...
+class MappedGraph(Graph[ValueType]):
+    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None) -> None:
+        super().__init__(vertexes)
+        self.__name_to_vertex: Dict[str, Vertex[ValueType]] = {vertex.name: vertex for vertex in vertexes} if vertexes else {}
+
+    def get_vertex(self, name: str) -> Optional[Vertex[ValueType]]:
+        return self.__name_to_vertex.get(name)
+
+    def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
+        old_vertex: Optional[Vertex[ValueType]] = self.get_vertex(name)
```
eprbell, line 79: Maybe rename `old_vertex` to `existing_vertex`?

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -53,6 +67,31 @@ class AssetPairAndTimestamp(NamedTuple):
...
+        if old_vertex:
+            return old_vertex
+
+        new_vertex: Vertex[ValueType] = Vertex[ValueType](name=name)
+        super().add_vertex(new_vertex)
```
eprbell, line 84: I think we need to override `add_vertex()` too: it can call its super and then add the vertex to the dict. Then here we can just call `add_vertex()`.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -179,35 +224,28 @@ def _build_fiat_list(self) -> None:
...
-    def _add_fiat_edges_to_graph(self, graph: Dict[str, Dict[str, None]], markets: Dict[str, List[str]]) -> None:
+    def _add_fiat_edges_to_graph(self, graph: MappedGraph[str], markets: Dict[str, List[str]]) -> None:
         if not self.__fiat_list:
             self._build_fiat_list()
 
         for fiat in self.__fiat_list:
             to_fiat_list: Dict[str, None] = dict.fromkeys(self.__fiat_list.copy())
             del to_fiat_list[fiat]
-            if graph.get(fiat):
+            # We don't want to add a fiat vertex here because that would allow a double hop on fiat (eg. USD -> KRW -> JPY)
```
eprbell, line 234: Just curious: what's wrong with a double fiat hop?

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -158,12 +165,9 @@ def __init__(
         self.__exchange_markets: Dict[str, Dict[str, List[str]]] = {}
         self.__google_api_key: Optional[str] = google_api_key
         self.__exchange_locked: bool = exchange_locked if exchange_locked is not None else False
-
-        # TO BE IMPLEMENTED - graph and vertex classes to make this more understandable
-        # https://github.com/eprbell/dali-rp2/pull/53#discussion_r924056308
         self.__default_exchange: str = _DEFAULT_EXCHANGE if default_exchange is None else default_exchange
         self.__exchange_csv_reader: Dict[str, Any] = {}
-        self.__exchange_graphs: Dict[str, Dict[str, Dict[str, None]]] = {}
+        self.__exchange_graphs: Dict[str, MappedGraph[str]] = {}
```
eprbell, line 170: Can you add a comment describing this mapping from string to graph?

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -436,12 +404,12 @@ def find_historical_bar(self, from_asset: str, to_asset: str, timestamp: datetim
...
         return result
 
-    def _add_alternative_markets(self, current_graph: Dict[str, Dict[str, None]], current_markets: Dict[str, List[str]]) -> None:
+    def _add_alternative_markets(self, graph: MappedGraph[str], current_markets: Dict[str, List[str]]) -> None:
         for base_asset, quote_asset in _ALT_MARKET_BY_BASE_DICT.items():
             alt_market = base_asset + quote_asset
             alt_exchange_name = _ALT_MARKET_EXCHANGES_DICT[alt_market]
 
-            # TO BE IMPLEMENTED - Add alt market to the end of list if another exchange exists already
+            # TO BE IMPLEMENTED - Add markets to a priorityqueue inside MappedGraph to prioritize higher volume exchanges
```
eprbell, line 412: priority queue

eprbell (changes requested):
Very nice: is this already working with real data?

src/dali/abstract_pair_converter_plugin.py
macanudo527: That DOES make more sense.

src/dali/abstract_pair_converter_plugin.py
macanudo527: Well, first of all, our current data provider for fiat pricing allows us to convert from any fiat to any other fiat, so it is unnecessary. But, on top of that, we only currently have daily exchange rates and so I felt like we need to prioritize the more granular data (crypto exchange rates) for now.

In the future, if we have a different provider, we may need to double hop, but we will need a different graph-building function at that point, so we can revisit it then.

macanudo527:
I hope this is what you were thinking. I added a function `add_vertex_by_name()` since this class works with names.

macanudo527:
This is good to go with real-world data. I did spot a bug with BETH pricing #143 that I'll have to address later, but it existed before this update.

eprbell:
> I hope this is what you were thinking. I added a function `add_vertex_by_name()` since this class works with names.
> 
> I've been trying to make a full run off and on. I should be able to finish it in the next 12 hours or so.

What I meant was to override `add_vertex()`:
```
    def add_vertex(self, vertex: Vertex[ValueType]) -> None:
        super().add_vertex(vertex)
        self.__name_to_vertex[vertex.name] = new_vertex
```

The `add_vertex_by_name()` is not mandatory in my opinion, but if you find it useful in your code you can add it.

eprbell:
> This is good to go with real-world data. I did spot a bug with BETH pricing #143 that I'll have to address later, but it existed before this update.

Cool, I'll review soon: I've been a bit swamped lately.

macanudo527:
> > This is good to go with real-world data. I did spot a bug with BETH pricing #143 that I'll have to address later, but it existed before this update.
> 
> Cool, I'll review soon: I've been a bit swamped

> > I hope this is what you were thinking. I added a function `add_vertex_by_name()` since this class works with names.
> > I've been trying to make a full run off and on. I should be able to finish it in the next 12 hours or so.
> 
> What I meant was to override `add_vertex()`:
> 
> ```
>     def add_vertex(self, vertex: Vertex[ValueType]) -> None:
>         super().add_vertex(vertex)
>         self.__name_to_vertex[vertex.name] = new_vertex
> ```
> 
> The `add_vertex_by_name()` is not mandatory in my opinion, but if you find it useful in your code you can add it.

That actually makes more sense. I guess I wasn't thinking.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -268,20 +233,25 @@ def get_historic_bar_from_native_source(self, timestamp: datetime, from_asset: s
         # else:
         # Graph building goes here.
 
-        pricing_path: Optional[List[str]] = self._bfs_cyclic(current_graph, from_asset, to_asset)
+        if from_asset_vertex and to_asset_vertex:
+            pricing_path = current_graph.dijkstra(from_asset_vertex, to_asset_vertex, False)
+        else:
+            raise RP2RuntimeError(f"The asset {from_asset} or {to_asset} is missing from graph")
```
eprbell, line 239: Nit:
``` suggestion
        if not from_asset_vertex or not to_asset_vertex:
            raise RP2RuntimeError(f"The asset {from_asset} or {to_asset} is missing from graph")
        pricing_path = current_graph.dijkstra(from_asset_vertex, to_asset_vertex, False)
```

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -450,10 +419,8 @@ def _add_alternative_markets(self, current_graph: Dict[str, Dict[str, None]], cu
                 alt_exchange: Exchange = _EXCHANGE_DICT[alt_exchange_name]()
                 self.__exchanges[alt_exchange_name] = alt_exchange
 
-            if current_graph.get(base_asset) and (quote_asset not in current_graph[base_asset]):
-                current_graph[base_asset][quote_asset] = None
-            else:
-                current_graph[base_asset] = {quote_asset: None}
+            # If the key doesn't exist, the MappedGraph will create a vertex and add it to the graph
```
eprbell, line 422: Can you clarify in the comment what the key is?

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -474,10 +441,8 @@ def _add_exchange_to_memcache(self, exchange: str) -> None:
 
             # TO BE IMPLEMENTED - lazy build graph only if needed
 
-            # Add the quote asset to the graph if it isn't there already.
-            current_graph.setdefault(market[_BASE], {})[market[_QUOTE]] = None
-
-        self._prioritize_quote_assets(current_markets)
+            # If the key doesn't exist, the GraphVertexesDict will create a vertex and add it to the graph
```
eprbell, line 444: Ditto.

eprbell (changes requested):
Just nits, this is a really nice improvement!

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -179,35 +227,28 @@ def _build_fiat_list(self) -> None:
...
+            if graph.get_vertex(fiat):
                 for to_be_added_fiat in to_fiat_list:
-                    # add a pair if it doesn't exist
-                    if to_be_added_fiat not in graph[fiat]:
-                        graph[fiat][to_be_added_fiat] = None
-            else:
-                graph[fiat] = to_fiat_list
+                    graph.add_neighbor(
+                        fiat,
+                        to_be_added_fiat,
```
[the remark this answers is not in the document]
macanudo527: fiat to fiat is fine and necessary.
fiat to fiat to fiat is what I was trying to avoid, and that is what I was calling a double hop.

Basically, I don't want nodes from the fiat 'exchange' linking to each other. I only want them to link to the fiat on the crypto exchanges.

macanudo527:
Did something need to be added to this @eprbell ?

eprbell (approved):
LGTM!

PR #144, opened 2023-04-29
Provide value to log string in dali.plugin.pair_converter.csv.kraken

qwhelan:
Actually include `file_name` in log message indicating file is missing from Kraken Google Drive.


macanudo527:
That's a rather odd omission. Sorry about that.

eprbell (approved):

Issue #145, opened 2023-04-29
Weigh vertexes in Djikstra based on monthly volume for the CCXT pair converter


macanudo527:
Currently the Djikstra search doesn't take into account trading volume when calculating routes. This could lead to inaccurate pricing caused by low volume. I propose weights be based on monthly volume. This should provide enough granularity. I'm afraid if we use daily volume, the graph could take up too much memory and pulling the data could take too long.

These weights will actually be added to a priority queue of exchanges in `MappedGraph` and the current private variable that tracks this `self.__exchange_markets` will be removed.

eprbell:
A question on this approach. Suppose I run DaLI on Apr 14 2023 and I get some results with certain pricing information reflecting the current volume. Then I run DaLI again on Apr 14 2024 and at that time volume data has completely changed: will the volume change affect transactions before Apr 14 2023? The reason I ask is that we should strive to preserve historical output as much as possible. In other words one way to test DaLI is to run it periodically and compare the output against golden files: such tests should continue to work.

macanudo527:
I think weights for the current month would have to be based on the previous month volume. 

For example if you run Dali on April 14th and price a transaction for April 3rd the exchange with the most volume in March of that year would be used. 

Alternatively, it could just be disabled for the most recent current month in order to keep consistent results. 

Otherwise, the volume would be month by month and never change. The exchange with the highest volume for March of 2023 will still be the exchange with the highest volume for March of 2023 in April 2024. 

The market priority queue would be sorted by month if that makes sense. 

For example, we price a transaction from April 2022 and exchange A has more volume than exchange B for April 2022, we will use exchange A. The volume for exchange A will never change once it's completed so it will always be the same price.

The priority queue for markets will choose which exchange to use for a given vertex. 

We have similar problem with vertexes in general though. The volume of transactions for BTC -> EUR is actually less than BTC -> USDT on a lot of exchanges and some stable coins are more popular on different exchanges at different times, so the quote priority will have to be created based on monthly volume as well.

eprbell:
Thanks, that makes sense. I guess the only time this would break price continuity is if a new exchange gets added, which may have largest volume for certain intervals in the past. But I think that particular case is probably OK and hopefully it doesn't happen very often.

macanudo527:
[...]

Issue #143, opened 2023-04-17
Pricing for dates before asset is listed on Binance.com are inaccurate


macanudo527:
If Kraken (the current default exchange used for pricing) or another non-Binance.com exchange is used as the default pricing exchange, BETH is priced inaccurately. It seems to be locked to a certain date, always reporting a similar `spot_price` for staking rewards. For some reason, regular buys on other exchanges don't seem to be affected. 

I'm guessing this has something to do with an inaccurate date. The bug popped up after Kraken became the default exchange.

EDIT: This is actually caused by the relatively new BETH->USDT market returning incorrect pricing when asking for prices before the asset was listed.

macanudo527:
I've contacted Binance.com support to see if there is way to find out when something was listed. If there isn't we can implement a volume filter that should cut out most bogus data points. This would work by rejecting any returned bars under a certain volume, 

I was thinking of implementing something where Dijkstra weights are determined by monthly volume in the future, but it might be better to implement it now to solve this issue.

macanudo527:
#146 fixes this with one caveat. Since untradeable assets are indistinguishable from those that the market hasn't opened yet for, users must declare untradeable assets (e.g. airdropped assets that don't have a market yet) in the CCXT pair plugin in order for them to be priced with the first weekly candle of the market once it opens. Users will be warned when this happens.

PR #146, opened 2023-05-02
Add Optimization to Djikstra

### LINKED ISSUE #140: Implement aliases for assets on specific exchanges
Some exchanges use ISO tickers for crypto assets. For example, Kraken (and Bitmex) uses XBT for BTC in some places. [More details on XBT vs BTC](https://support.kraken.com/hc/en-us/articles/360001206766-Bitcoin-currency-code-XBT-vs-BTC)

Other exchanges use 'micro' tickers for certain assets to denote fractions of a crypto asset. For example, Pionex uses mBTC to refer to 0.001 BTC.

In order to price these appropriately, we need to implement an alias system that will create a vector between the appropriate asset the alias represents and have a locked `spot_price` of 1 if they are a true alias or a fraction if they represent a micro asset. These will then be added to the Djikstra graph with 0 weight.

This helps address #133 and issues that came up with #114

macanudo527:
This addresses #145 , #143 and helps clear the way to resolve #140 

This will allow markets (or edges in the Djikstra search) to be added and deleted when cloning a graph. It also allows the optimization of a graph by altering the weights of edges during the cloning.

This is **WIP**. I submitted it to get feedback on the `MappedGraph` data class before implementing it in the CCXT pair converter.

To-do:
- [x] Write tests for optimization
- [x] Modify `_add_exchange_to_memcache` to pass unoptimized graph, and rename it `_generate_unoptimized_graph`
- [x] Modify CSV reader to return weekly candles
- [x] Add function to optimize an asset's weights in `ccxt.py`
- [x] Add check for optimization before REST call for price
- [x] Add validation of bar timestamp in order to find markets that are not available
- [x] Modify cloning function to receive multiple assets
- [x] Create manifest from transactions and pass it on to the pair converter plugins if needed
- [x] Add snapshot compression to reduce excess snapshots
- [x] Clean code

### DIFF
--- a/src/dali/abstract_pair_converter_plugin.py
+++ b/src/dali/abstract_pair_converter_plugin.py
@@ -24,7 +24,7 @@ from requests.exceptions import ReadTimeout
 from requests.models import Response
 from requests.sessions import Session
 from rp2.rp2_decimal import ZERO, RP2Decimal
-from rp2.rp2_error import RP2RuntimeError, RP2TypeError
+from rp2.rp2_error import RP2RuntimeError, RP2TypeError, RP2ValueError
 
 from dali.cache import load_from_cache, save_to_cache
 from dali.configuration import HISTORICAL_PRICE_KEYWORD_SET
@@ -68,9 +68,14 @@ class AssetPairAndTimestamp(NamedTuple):
 
 
 class MappedGraph(Graph[ValueType]):
-    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None) -> None:
+    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None, optimized_assets: Optional[Dict[str, None]] = None) -> None:
         super().__init__(vertexes)
         self.__name_to_vertex: Dict[str, Vertex[ValueType]] = {vertex.name: vertex for vertex in vertexes} if vertexes else {}
+        self.__optimized_assets: Dict[str, None] = {} if optimized_assets is None else optimized_assets
+
+    def add_missing_vertex(self, name: str) -> None:
+        if not self.__name_to_vertex.get(name):
+            self.add_vertex(Vertex[ValueType](name=name))
 
     def add_vertex(self, vertex: Vertex[ValueType]) -> None:
         super().add_vertex(vertex)
@@ -87,6 +92,45 @@ class MappedGraph(Graph[ValueType]):
         new_vertex: Vertex[ValueType] = Vertex[ValueType](name=name)
         self.add_vertex(new_vertex)
         return new_vertex
+
+    def is_optimized(self, asset: str) -> bool:
+        return asset in self.__optimized_assets
+
+    # Optimized asset is the asset we are optimizing during this cloning.
+    # Optimized assets are tracked to prevent redundant cloning.
+    # Negative weights will get deleted.
+    def clone_with_optimization(self, optimized_asset: str, neighbor_weights: Dict[str, float]) -> "MappedGraph[ValueType]":
+        if optimized_asset in self.__optimized_assets:
+            raise RP2ValueError(
+                "You are trying to optimize an asset that has already been optimized."
+                "Use is_optimized() to check if an asset has already been optimized before cloning."
+            )
+
+        cloned_optimized_assets: Dict[str, None] = self.__optimized_assets
+        cloned_optimized_assets[optimized_asset] = None
+        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=cloned_optimized_assets)
+
+        for original_vertex in self.vertexes:
+            # Add existing neighbors
+            for neighbor in original_vertex.neighbors:
+                neighbor_weight: float
+                if original_vertex.name == optimized_asset:
+                    neighbor_weight = neighbor_weights.pop(neighbor.name, original_vertex.get_weight(neighbor))
+                else:
+                    neighbor_weight = original_vertex.get_weight(neighbor)
+
+                # Delete neighbor if negative weight
+                if neighbor_weight >= 0.0:
+                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor.name, neighbor_weight)
+                else:
+                    cloned_mapped_graph.add_missing_vertex(original_vertex.name)
+
+            # Add new neighbors
+            for neighbor_name in neighbor_weights.keys():
+                if original_vertex.name == optimized_asset:
+                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor_name, neighbor_weights[neighbor_name])
+
+        return cloned_mapped_graph
 
     def add_neighbor(self, vertex_name: str, neighbor_name: str, weight: float = 0.0) -> None:
         vertex: Vertex[ValueType] = self.get_or_set_vertex(vertex_name)
--- a/src/dali/abstract_pair_converter_plugin.py
+++ b/src/dali/abstract_pair_converter_plugin.py
@@ -14,7 +14,7 @@
 
 from datetime import datetime, timedelta
 from json import JSONDecodeError
-from typing import Any, Dict, List, NamedTuple, Optional, cast
+from typing import Any, Dict, List, NamedTuple, Optional, Set, cast
 
 import requests
 from prezzemolo.graph import Graph
@@ -67,24 +67,40 @@ class AssetPairAndTimestamp(NamedTuple):
     exchange: str
 
 
+class TransactionManifest(NamedTuple):
+    assets: Set[str]
+    exchanges: Set[str]
+    first_transaction_datetime: datetime
+
+
 class MappedGraph(Graph[ValueType]):
-    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None, optimized_assets: Optional[Dict[str, None]] = None) -> None:
+    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None, optimized_assets: Optional[Set[str]] = None) -> None:
         super().__init__(vertexes)
         self.__name_to_vertex: Dict[str, Vertex[ValueType]] = {vertex.name: vertex for vertex in vertexes} if vertexes else {}
-        self.__optimized_assets: Dict[str, None] = {} if optimized_assets is None else optimized_assets
+        self.__optimized_assets: Set[str] = set() if optimized_assets is None else optimized_assets
 
-    def add_missing_vertex(self, name: str) -> None:
+    def add_vertex_if_missing(self, name: str) -> None:
         if not self.__name_to_vertex.get(name):
             self.add_vertex(Vertex[ValueType](name=name))
+
+    def get_all_children_of_vertex(self, vertex: Vertex[ValueType]) -> Set[Vertex[ValueType]]:
+        children = set(vertex.neighbors)
+        for neighbor in vertex.neighbors:
+            children.update(self.get_all_children_of_vertex(neighbor) or set())
+        return children
 
     def add_vertex(self, vertex: Vertex[ValueType]) -> None:
         super().add_vertex(vertex)
         self.__name_to_vertex[vertex.name] = vertex
 
     def get_vertex(self, name: str) -> Optional[Vertex[ValueType]]:
+        if not isinstance(name, str):
+            raise RP2TypeError(f"Internal Error: parameter {name} is not a str.")
         return self.__name_to_vertex.get(name)
 
     def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
+        if not isinstance(name, str):
+            raise RP2TypeError(f"Internal Error: parameter {name} is not a str.")
         existing_vertex: Optional[Vertex[ValueType]] = self.get_vertex(name)
         if existing_vertex:
             return existing_vertex
@@ -96,47 +112,51 @@ class MappedGraph(Graph[ValueType]):
     def is_optimized(self, asset: str) -> bool:
         return asset in self.__optimized_assets
 
-    # Optimized asset is the asset we are optimizing during this cloning.
-    # Optimized assets are tracked to prevent redundant cloning.
-    # Negative weights will get deleted.
-    def clone_with_optimization(self, optimized_asset: str, neighbor_weights: Dict[str, float]) -> "MappedGraph[ValueType]":
-        if optimized_asset in self.__optimized_assets:
-            raise RP2ValueError(
-                "You are trying to optimize an asset that has already been optimized."
-                "Use is_optimized() to check if an asset has already been optimized before cloning."
-            )
+    @property
+    def optimized_assets(self) -> Set[str]:
+        return self.__optimized_assets
 
-        cloned_optimized_assets: Dict[str, None] = self.__optimized_assets
-        cloned_optimized_assets[optimized_asset] = None
-        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=cloned_optimized_assets)
+    # Optimization contains a dict with a key of the optimized asset and a value of a dict with the optimized weights for each neighbor
+    # Optimized assets are tracked to prevent requesting prices for unoptimized assets
+    # Negative weights will get deleted.
+    def clone_with_optimization(self, optimization: Dict[str, Dict[str, float]]) -> "MappedGraph[ValueType]":
+
+        # wrapping optimized_assets in set() makes a deep copy
+        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=set(self.optimized_assets))
 
         for original_vertex in self.vertexes:
             # Add existing neighbors
             for neighbor in original_vertex.neighbors:
                 neighbor_weight: float
-                if original_vertex.name == optimized_asset:
-                    neighbor_weight = neighbor_weights.pop(neighbor.name, original_vertex.get_weight(neighbor))
+                optimized: bool = False
+                if original_vertex.name in set(optimization.keys()):
+                    neighbor_weight = optimization[original_vertex.name].pop(neighbor.name, original_vertex.get_weight(neighbor))
+                    optimized = True
                 else:
                     neighbor_weight = original_vertex.get_weight(neighbor)
 
                 # Delete neighbor if negative weight
                 if neighbor_weight >= 0.0:
-                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor.name, neighbor_weight)
+                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor.name, neighbor_weight, optimized)
                 else:
-                    cloned_mapped_graph.add_missing_vertex(original_vertex.name)
+                    cloned_mapped_graph.add_vertex_if_missing(original_vertex.name)
 
             # Add new neighbors
-            for neighbor_name in neighbor_weights.keys():
-                if original_vertex.name == optimized_asset:
-                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor_name, neighbor_weights[neighbor_name])
+            for optimized_asset, neighbor_weights in optimization.items():
+                for neighbor_name in neighbor_weights.keys():
+                    if original_vertex.name == optimized_asset:
+                        cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor_name, neighbor_weights[neighbor_name])
 
         return cloned_mapped_graph
 
-    def add_neighbor(self, vertex_name: str, neighbor_name: str, weight: float = 0.0) -> None:
+    # Marking weights as optimized prevents REST API calls to re-optimize them
+    def add_neighbor(self, vertex_name: str, neighbor_name: str, weight: float = 0.0, optimized: bool = False) -> None:
         vertex: Vertex[ValueType] = self.get_or_set_vertex(vertex_name)
         neighbor: Vertex[ValueType] = self.get_or_set_vertex(neighbor_name)
         if not vertex.has_neighbor(neighbor):
             vertex.add_neighbor(neighbor, weight)
+        if optimized:
+            self.__optimized_assets.add(vertex_name)
 
 
 class AbstractPairConverterPlugin:
@@ -150,11 +170,7 @@ class AbstractPairConverterPlugin:
             raise RP2TypeError(
                 f"historical_price_type must be one of {', '.join(sorted(HISTORICAL_PRICE_KEYWORD_SET))}, instead it was: {historical_price_type}"
             )
-        try:
-            result = cast(Dict[AssetPairAndTimestamp, HistoricalBar], load_from_cache(self.cache_key()))
-        except EOFError:
-            LOGGER.error("EOFError: Cached file corrupted, no cache found.")
-            result = None
+        result = cast(Dict[AssetPairAndTimestamp, HistoricalBar], load_from_cache(self.cache_key()))
         self.__cache: Dict[AssetPairAndTimestamp, HistoricalBar] = result if result is not None else {}
         self.__historical_price_type: str = historical_price_type
         self.__session: Session = requests.Session()
@@ -172,6 +188,9 @@ class AbstractPairConverterPlugin:
         raise NotImplementedError("Abstract method: it must be implemented in the plugin class")
 
     def cache_key(self) -> str:
+        raise NotImplementedError("Abstract method: it must be implemented in the plugin class")
+
+    def optimize(self, transaction_manifest: TransactionManifest) -> None:
         raise NotImplementedError("Abstract method: it must be implemented in the plugin class")
 
     def _add_bar_to_cache(self, key: AssetPairAndTimestamp, historical_bar: HistoricalBar) -> None:
@@ -289,6 +308,7 @@ class AbstractPairConverterPlugin:
                         fiat,
                         to_be_added_fiat,
                         self.__fiat_priority.get(fiat, _STANDARD_WEIGHT),
+                        True,  # use set optimization
                     )
 
                 LOGGER.debug("Added to assets for %s: %s", fiat, to_fiat_list)
--- a/src/dali/abstract_pair_converter_plugin.py
+++ b/src/dali/abstract_pair_converter_plugin.py
@@ -120,7 +120,6 @@ class MappedGraph(Graph[ValueType]):
     # Optimized assets are tracked to prevent requesting prices for unoptimized assets
     # Negative weights will get deleted.
     def clone_with_optimization(self, optimization: Dict[str, Dict[str, float]]) -> "MappedGraph[ValueType]":
-
         cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=self.__optimized_assets.copy())
 
         for original_vertex in self.vertexes:
--- a/src/dali/abstract_pair_converter_plugin.py
+++ b/src/dali/abstract_pair_converter_plugin.py
@@ -24,12 +24,13 @@ from requests.exceptions import ReadTimeout
 from requests.models import Response
 from requests.sessions import Session
 from rp2.rp2_decimal import ZERO, RP2Decimal
-from rp2.rp2_error import RP2RuntimeError, RP2TypeError, RP2ValueError
+from rp2.rp2_error import RP2RuntimeError, RP2TypeError
 
 from dali.cache import load_from_cache, save_to_cache
 from dali.configuration import HISTORICAL_PRICE_KEYWORD_SET
 from dali.historical_bar import HistoricalBar
 from dali.logger import LOGGER
+from dali.transaction_manifest import TransactionManifest
 
 # exchangerates.host keywords
 _SUCCESS: str = "success"
@@ -67,12 +68,6 @@ class AssetPairAndTimestamp(NamedTuple):
     exchange: str
 
 
-class TransactionManifest(NamedTuple):
-    assets: Set[str]
-    exchanges: Set[str]
-    first_transaction_datetime: datetime
-
-
 class MappedGraph(Graph[ValueType]):
     def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None, optimized_assets: Optional[Set[str]] = None) -> None:
         super().__init__(vertexes)
@@ -83,10 +78,14 @@ class MappedGraph(Graph[ValueType]):
         if not self.__name_to_vertex.get(name):
             self.add_vertex(Vertex[ValueType](name=name))
 
-    def get_all_children_of_vertex(self, vertex: Vertex[ValueType]) -> Set[Vertex[ValueType]]:
+    def get_all_children_of_vertex(self, vertex: Vertex[ValueType], visited: Optional[Set[Vertex[ValueType]]] = None) -> Set[Vertex[ValueType]]:
+        # We need to keep track of the visited vertexes to prevent infinite recursion
+        visited = set() if visited is None else visited
         children = set(vertex.neighbors)
+        visited.add(vertex)
         for neighbor in vertex.neighbors:
-            children.update(self.get_all_children_of_vertex(neighbor) or set())
+            if neighbor not in visited:
+                children.update(self.get_all_children_of_vertex(neighbor, visited))
         return children
 
     def add_vertex(self, vertex: Vertex[ValueType]) -> None:
@@ -110,7 +109,8 @@ class MappedGraph(Graph[ValueType]):
         return new_vertex
 
     def is_optimized(self, asset: str) -> bool:
-        return asset in self.__optimized_assets
+        LOGGER.debug("Checking if %s is in %s", asset, self.__optimized_assets)
+        return bool(asset in self.__optimized_assets)
 
     @property
     def optimized_assets(self) -> Set[str]:
@@ -121,8 +121,7 @@ class MappedGraph(Graph[ValueType]):
     # Negative weights will get deleted.
     def clone_with_optimization(self, optimization: Dict[str, Dict[str, float]]) -> "MappedGraph[ValueType]":
 
-        # wrapping optimized_assets in set() makes a deep copy
-        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=set(self.optimized_assets))
+        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=self.__optimized_assets.copy())
 
         for original_vertex in self.vertexes:
             # Add existing neighbors
@@ -170,8 +169,12 @@ class AbstractPairConverterPlugin:
             raise RP2TypeError(
                 f"historical_price_type must be one of {', '.join(sorted(HISTORICAL_PRICE_KEYWORD_SET))}, instead it was: {historical_price_type}"
             )
-        result = cast(Dict[AssetPairAndTimestamp, HistoricalBar], load_from_cache(self.cache_key()))
-        self.__cache: Dict[AssetPairAndTimestamp, HistoricalBar] = result if result is not None else {}
+        try:
+            result = cast(Dict[AssetPairAndTimestamp, HistoricalBar], load_from_cache(self.cache_key()))
+        except EOFError:
+            LOGGER.error("EOFError: Cached file corrupted, no cache found.")
+            result = None
+        self.__cache: Dict[AssetPairAndTimestamp, Any] = result if result is not None else {}
         self.__historical_price_type: str = historical_price_type
         self.__session: Session = requests.Session()
         self.__fiat_list: List[str] = []
@@ -198,6 +201,13 @@ class AbstractPairConverterPlugin:
 
     def _get_bar_from_cache(self, key: AssetPairAndTimestamp) -> Optional[HistoricalBar]:
         return self.__cache.get(self._floor_key(key))
+
+    # All bundle timestamps have 1 millisecond added to them, so will not conflict with the floored timestamps of single bars
+    def _add_bundle_to_cache(self, key: AssetPairAndTimestamp, historical_bars: List[HistoricalBar]) -> None:
+        self.__cache[key] = historical_bars
+
+    def _get_bundle_from_cache(self, key: AssetPairAndTimestamp) -> Optional[List[HistoricalBar]]:
+        return cast(List[HistoricalBar], self.__cache.get(key))
 
     # The most granular pricing available is 1 minute, to reduce the size of cache and increase the reuse of pricing data
     def _floor_key(self, key: AssetPairAndTimestamp) -> AssetPairAndTimestamp:
--- a/src/dali/dali_main.py
+++ b/src/dali/dali_main.py
@@ -155,8 +155,18 @@ def _dali_main_internal(country: AbstractCountry) -> None:
             pair_converter_list.append(CcxtPairConverterPlugin(Keyword.HISTORICAL_PRICE_HIGH.value))
             LOGGER.info("No pair converter plugins found in configuration file: using default pair converters.")
 
+        with ThreadPool(args.thread_count) as pool:
+            result_list = pool.map(_input_plugin_helper, input_plugin_args_list)
+
+        transactions: List[AbstractTransaction] = [transaction for result in result_list for transaction in result]
+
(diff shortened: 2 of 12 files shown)


src/dali/abstract_pair_converter_plugin.py
```diff
@@ -68,9 +68,14 @@ class AssetPairAndTimestamp(NamedTuple):
 
 
 class MappedGraph(Graph[ValueType]):
-    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None) -> None:
+    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None, optimized_assets: Optional[Dict[str, None]] = None) -> None:
```
eprbell, line 71: I have a few questions:
* I'm not clear why assets are qualified as "optimized": could you clarify the meaning of the term?
* why does init receive an `optimized_assets` parameter? The way I was thinking of it was that init would keep the same signature and clone would receive `changed_edges`, `new_edges` and `deleted_edges`. Would that work or am I missing something?

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -68,9 +68,14 @@ class AssetPairAndTimestamp(NamedTuple):
...
         super().__init__(vertexes)
         self.__name_to_vertex: Dict[str, Vertex[ValueType]] = {vertex.name: vertex for vertex in vertexes} if vertexes else {}
+        self.__optimized_assets: Dict[str, None] = {} if optimized_assets is None else optimized_assets
+
+    def add_missing_vertex(self, name: str) -> None:
```
eprbell, line 76: I'd call this `add_vertex_by_name`.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,6 +93,45 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+            raise RP2ValueError(
+                "You are trying to optimize an asset that has already been optimized."
+                "Use is_optimized() to check if an asset has already been optimized before cloning."
+            )
+
+        cloned_optimized_assets: Dict[str, None] = self.__optimized_assets
+        cloned_optimized_assets[optimized_asset] = None
+        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=cloned_optimized_assets)
+
+        for original_vertex in super().vertexes:
```
eprbell, line 113: No need for `super()`: you can just use `self`.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,6 +93,45 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+            # Add existing neighbors
+            for neighbor in original_vertex.neighbors:
+                neighbor_weight: float
+                if original_vertex.name == optimized_asset:
```
eprbell, line 117: I'm not too clear what this bit is doing (it may be caused by my original uncertainty with the "optimized" concept)

eprbell (commented):
Here are some initial comments. BTW, thanks for working on this complex functionality: really appreciate you tackling this hard problem!

src/dali/abstract_pair_converter_plugin.py
macanudo527: It's probably best illustrated with an example,

I have two transactions for BTC in one month.

The CCXT pair converter goes to price the first transaction, finds it's un-optimized by checking using `is_optimized()` and proceeds to optimize the graph. It makes ~5 REST API calls to find the quote asset with the most volume, calculates new weights and clones the graph. It adds BTC to the `self.__optimized_assets` dictionary when it was cloned. Now the new clone has that asset marked as optimized.

We get to the second transaction that month for BTC. CCXT pair converter checks and finds BTC is optimized and it doesn't have to make the costly REST API calls, saving time.

In an ideal world, we would optimize the entire graph, but of course that would take a prohibitively long time, so we optimize just what is needed and carry that over into the new graph. 

Later, when we pull an ETH transaction, we will optimize that and copy over the optimization into the new graph, so that if we encounter another BTC transaction during that month the optimization is still there for BTC and we don't need to make unnecessary REST API calls.

src/dali/abstract_pair_converter_plugin.py
macanudo527: We are only going to be optimizing one asset at a time. So, we need to change the weights of its neighbors (ie. quote assets) only if they were passed to the function. If they were not passed to the function, we keep the original weights.

macanudo527:
I'm afraid there has been a misunderstanding here because it doesn't look like I can implement this using `AVLTree` from prezzemolo. I will need to replace nodes in the tree, and it appears the AVL tree is immutable.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,6 +93,45 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+                # Delete neighbor if negative weight
+                if neighbor_weight >= 0.0:
+                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor.name, neighbor_weight)
+                else:
+                    cloned_mapped_graph.add_missing_vertex(original_vertex.name)
+
+            # Add new neighbors
+            for neighbor_name in neighbor_weights.keys():
+                if original_vertex.name == optimized_asset:
+                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor_name, neighbor_weights[neighbor_name])
```
eprbell, line 131: Could the neighbors be already added? If not can you clarify in a comment why?

eprbell (commented):
Ok, I understand this much better, thanks for all the explanation. Some more comments inline.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,6 +93,45 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+                if original_vertex.name == optimized_asset:
+                    neighbor_weight = neighbor_weights.pop(neighbor.name, original_vertex.get_weight(neighbor))
+                else:
+                    neighbor_weight = original_vertex.get_weight(neighbor)
+
+                # Delete neighbor if negative weight
+                if neighbor_weight >= 0.0:
+                    cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor.name, neighbor_weight)
+                else:
+                    cloned_mapped_graph.add_missing_vertex(original_vertex.name)
```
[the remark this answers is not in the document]
macanudo527: I don't see the need for negative weights at this time. I can't imagine a use case. I can see where it would be useful to make edge deletion explicit though. I was just trying to reduce the number of variables that need to be shuffled around.

src/dali/abstract_pair_converter_plugin.py
macanudo527: The `add_neighbor()` function is overwritten to prevent adding a neighbor that has already been added.

macanudo527:
@eprbell Here it is, my thesis project.

I tried my best to cleanly cut out each feature into its own commit to make this easier to review. If you find it to unwieldy, let me know I can try to separate it out more or make another PR, but a lot of this was very interrelated.

eprbell:
> > @eprbell Here it is, my thesis project.
> > I tried my best to cleanly cut out each feature into its own commit to make this easier to review. If you find it to unwieldy, let me know I can try to separate it out more or make another PR, but a lot of this was very interrelated.
> 
> Awesome! I'm traveling internationally at the moment, so it may take a bit longer for me to review it. I'll get to it in the next few days.

Sorry for being late in reviewing this. I'm on vacation until the 19th, traveling in southern Europe. I was hoping to get some time to review PRs but it didn't happen. I'll get to it as soon as I can, but it may be slower than usual. Thanks for your patience, Neal!

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -67,19 +67,40 @@ class AssetPairAndTimestamp(NamedTuple):
...
+        self.__optimized_assets: Set[str] = set() if optimized_assets is None else optimized_assets
+
+    def add_vertex_if_missing(self, name: str) -> None:
+        if not self.__name_to_vertex.get(name):
+            self.add_vertex(Vertex[ValueType](name=name))
+
+    def get_all_children_of_vertex(self, vertex: Vertex[ValueType]) -> Set[Vertex[ValueType]]:
+        children = set(vertex.neighbors)
+        for neighbor in vertex.neighbors:
+            children.update(self.get_all_children_of_vertex(neighbor) or set())
```
eprbell, line 89: Why the or?

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,11 +109,61 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+        return self.__optimized_assets
+
+    # Optimization contains a dict with a key of the optimized asset and a value of a dict with the optimized weights for each neighbor
+    # Optimized assets are tracked to prevent requesting prices for unoptimized assets
+    # Negative weights will get deleted.
+    def clone_with_optimization(self, optimization: Dict[str, Dict[str, float]]) -> "MappedGraph[ValueType]":
+        for optimized_asset in optimization.keys():
+            if optimized_asset in self.__optimized_assets:
+                raise RP2ValueError(
+                    f"INTERNAL ERROR: You are trying to optimize an asset that has already been optimized. "
```
eprbell, line 126: Use impersonal form (I.e.: "attempt to optimize..." ): "you" seems directed to the user, but this is an internal error and the user has no way to fix this

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,11 +109,61 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
 
-    def add_neighbor(self, vertex_name: str, neighbor_name: str, weight: float = 0.0) -> None:
+    def is_optimized(self, asset: str) -> bool:
+        return asset in self.__optimized_assets
+
+    @property
+    def optimized_assets(self) -> Set[str]:
+        return self.__optimized_assets
+
+    # Optimization contains a dict with a key of the optimized asset and a value of a dict with the optimized weights for each neighbor
```
eprbell, line 119: Can optimizations be updated? Can you add a description of how that works?

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,11 +109,61 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+
+        # wrapping optimized_assets in set() makes a deep copy
+        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=set(self.optimized_assets))
+
+        for original_vertex in self.vertexes:
+            # Add existing neighbors
+            for neighbor in original_vertex.neighbors:
+                neighbor_weight: float
+                optimized: bool = False
+                if original_vertex.name in set(optimization.keys()):
```
eprbell, line 139: Assign the set to a variable outside of the loop, to avoid rebuilding it at every iteration.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -106,11 +177,7 @@ def __init__(self, historical_price_type: str, fiat_priority: Optional[str] = No
             raise RP2TypeError(
                 f"historical_price_type must be one of {', '.join(sorted(HISTORICAL_PRICE_KEYWORD_SET))}, instead it was: {historical_price_type}"
             )
-        try:
-            result = cast(Dict[AssetPairAndTimestamp, HistoricalBar], load_from_cache(self.cache_key()))
-        except EOFError:
-            LOGGER.error("EOFError: Cached file corrupted, no cache found.")
-            result = None
+        result = cast(Dict[AssetPairAndTimestamp, HistoricalBar], load_from_cache(self.cache_key()))
```
eprbell, line 180: Curious: why is the try/catch no longer needed?

src/dali/manifest_builder.py
```diff
@@ -0,0 +1,64 @@
...
+from multiprocessing.pool import ThreadPool
+from typing import List, Set
+
+from dali.abstract_pair_converter_plugin import TransactionManifest
+from dali.abstract_transaction import AbstractTransaction
+from dali.in_transaction import InTransaction
+from dali.intra_transaction import IntraTransaction
+from dali.out_transaction import OutTransaction
+
+
```
eprbell, line 25: Can you add a comment describing what the manifest is, how it's used and what it contains?

eprbell (changes requested):
This review is incomplete: I'll give you feedback in chunks as I go through all the code over the next few days.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -67,19 +67,40 @@ class AssetPairAndTimestamp(NamedTuple):
...
+    def __init__(self, vertexes: Optional[List["Vertex[ValueType]"]] = None, optimized_assets: Optional[Set[str]] = None) -> None:
         super().__init__(vertexes)
         self.__name_to_vertex: Dict[str, Vertex[ValueType]] = {vertex.name: vertex for vertex in vertexes} if vertexes else {}
+        self.__optimized_assets: Set[str] = set() if optimized_assets is None else optimized_assets
+
+    def add_vertex_if_missing(self, name: str) -> None:
+        if not self.__name_to_vertex.get(name):
+            self.add_vertex(Vertex[ValueType](name=name))
+
+    def get_all_children_of_vertex(self, vertex: Vertex[ValueType]) -> Set[Vertex[ValueType]]:
```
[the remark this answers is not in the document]
macanudo527: The graphs being built don't loop back around. They only move toward the quote asset. Having said that though, we should include some kind of safety since this could infinitely recurse under some unforeseen situation.

src/dali/abstract_pair_converter_plugin.py
macanudo527: Optimizations are only needed for the cloning process. They are basically immutable, but at line 686 of `ccxt.py` the volumes are changed to weights.

We talked earlier about keeping things immutable and I attempted that here. I guess to make it more immutable, the function to add an optimized neighbor `add_neighbor()` needs to be made private.

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -88,11 +109,61 @@ def get_or_set_vertex(self, name: str) -> Vertex[ValueType]:
...
+        for optimized_asset in optimization.keys():
+            if optimized_asset in self.__optimized_assets:
+                raise RP2ValueError(
+                    f"INTERNAL ERROR: You are trying to optimize an asset that has already been optimized. "
+                    f"Use is_optimized() to check if an asset has already been optimized before cloning."
+                    f"Found {optimized_asset} in {self.__optimized_assets}."
+                )
+
+        # wrapping optimized_assets in set() makes a deep copy
+        cloned_mapped_graph: MappedGraph[ValueType] = MappedGraph(optimized_assets=set(self.optimized_assets))
```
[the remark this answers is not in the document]
macanudo527: `deepcopy()` right? I want to cut all the references.

src/dali/abstract_pair_converter_plugin.py
macanudo527: This looks like a rebase error. I'm still mastering my rebase skills and sublime merge.

macanudo527:
Don't worry about the delay in reviewing. I'm quite busy with another project and job-hunting. We will chip away at it.

src/dali/abstract_pair_converter_plugin.py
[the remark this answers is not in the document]
eprbell: Agreed, we should throw an internal error if a cycle is detected.

src/dali/abstract_pair_converter_plugin.py
eprbell: Immutability is fine but I was thinking of the case in which an optimization might become outdated (e.g. volume data changes, a market disappears and a new route is needed, etc.). Is there a way to recompute optimizations is such a case (this may require to clone a new graph, but that's fine). In other words: how final is it when an asset appears in __optimized_assets?

eprbell:
> Don't worry about the delay in reviewing. I'm quite busy with another project and job-hunting. We will chip away at it.

Sounds good! Good luck with the job hunt, and if you need a reference just let me know on private channel.

src/dali/dali_main.py
```diff
@@ -154,8 +158,18 @@ def _dali_main_internal(country: AbstractCountry) -> None:
             pair_converter_list.append(CcxtPairConverterPlugin(Keyword.HISTORICAL_PRICE_HIGH.value))
             LOGGER.info("No pair converter plugins found in configuration file: using default pair converters.")
 
+        with ThreadPool(args.thread_count) as pool:
+            result_list = pool.map(_input_plugin_helper, input_plugin_args_list)
+
+        transactions: List[AbstractTransaction] = [transaction for result in result_list for transaction in result]
+
+        LOGGER.info("Building manifest for optimization")
```
eprbell, line 166: Can you make the message a little more specific? E.g. Manifest for price optimization, or something like that.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -210,9 +217,12 @@ def __init__(
         self.__exchange_locked: bool = exchange_locked if exchange_locked is not None else False
         self.__default_exchange: str = _DEFAULT_EXCHANGE if default_exchange is None else default_exchange
         self.__exchange_csv_reader: Dict[str, Any] = {}
-        # key: name of exchange, value: graph that prioritizes that exchange
-        self.__exchange_graphs: Dict[str, MappedGraph[str]] = {}
+        # key: name of exchange, value: AVLTree of all snapshots of the graph
+        # TO BE IMPLEMENTED - Combine all graphs into one graph where assets can 'teleport' between exchanges
+        #   This will eliminate the need for markets and this dict, replacing it with just one AVLTree
+        self.__exchange_datetime_graph_tree_dict: Dict[str, AVLTree[datetime, MappedGraph[str]]] = {}
```
eprbell, line 223: I suggest renaming this to __exchange_2_graph_tree or similar.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -251,17 +264,22 @@ def get_historic_bar_from_native_source(self, timestamp: datetime, from_asset: s
...
             else:
                 self.__logger.error("WARNING: Unrecognized Exchange: %s. Please open an issue at %s", exchange, self.issues_url)
                 return None
 
         current_markets = self.__exchange_markets[exchange]
-        current_graph = self.__exchange_graphs[exchange]
+        current_graph = self.__exchange_datetime_graph_tree_dict[exchange].find_max_value_less_than(timestamp)
+        if current_graph is None:
+            raise RP2RuntimeError(
+                "Internal Error: Graph snapshot doesn't exist. Either you are trying to route a price before "
```
eprbell, line 280: Use impersonal form with internal errors.

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -169,72 +171,135 @@ def _split_chunks_size_n(self, file_name: str, csv_file: str, chunk_size: int =
...
+                    # The timestamp of the first row becomes the timestamp for the weekly row
+                    col_sums: List[str] = [chunk[0][self.__TIMESTAMP_INDEX]]
+
+                    # We don't want/need to add up the timestamp column
+                    for col in range(self.__OPEN, self.__TRADES + 1):
+                        col_sum = str(sum((RP2Decimal(row[col]) for row in seven_chunks), ZERO))
+
+                        # Average all prices
+                        if col in range(self.__OPEN, (self.__CLOSE + 1)):
+                            col_sum = str(RP2Decimal(col_sum) / RP2Decimal("7"))
```
eprbell, line 195: You may want to use a different variable to store the average, e.g. column_average

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -169,72 +171,135 @@ def _split_chunks_size_n(self, file_name: str, csv_file: str, chunk_size: int =
...
+                week_chunk = []
+                for i in range(0, len(chunk), 7):
+                    seven_chunks = chunk[i : i + 7]
+
+                    # We want to make sure we have a full week for the averages and accurate pricing
+                    if len(seven_chunks) < 7:
+                        break
+
+                    # The timestamp of the first row becomes the timestamp for the weekly row
+                    col_sums: List[str] = [chunk[0][self.__TIMESTAMP_INDEX]]
```
eprbell, line 187: Use full names: column_sums

eprbell (changes requested):
Here's another round.

src/dali/abstract_pair_converter_plugin.py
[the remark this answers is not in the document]
macanudo527: Ok, I'll test it. It was causing some hard to debug errors.

src/dali/abstract_pair_converter_plugin.py
macanudo527: I think there is a misunderstanding about how the optimization process works. Currently, the CCXT plugin uses the manifest to walk through week by week the volume of the exchanges and weighing the paths based on volume. These weeks are then joined together if they have the same ranking.

For example, we optimize for BTC -> USD on an exchange that doesn't have USD, so we must route through a stable coin:

week 1
BTC -> USDT volume is 100
BTC -> USDC volume is 10
week 2
BTC -> USDT volume is 100
BTC -> USDC volume is 20
week 3
BTC -> USDT volume is 10
BTC -> USDC volume is 200

When a transaction comes up requesting the price of BTC, CCXT will begin to optimize the graph snapshots for that exchange. It will make a snapshot of the graph for week 1 and 2 and place it in the AVLTree with a timestamp for week 1. There is no snapshot for week 2 since it would be identical. A new snapshot __is__ made for week 3 and placed in the AVLTree at week 3 timestamp.

Once these snapshots are made, currently, they are not changed or edited. They are used for the entire pricing process. Since the past can't change, there is no need to reoptimize a graph snapshot once it is made. Does that make sense?

src/dali/abstract_pair_converter_plugin.py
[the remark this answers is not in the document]
macanudo527: This does actually lead to infinite recursion with real-world data. I'll have to rewrite it to prevent that.

src/dali/abstract_pair_converter_plugin.py
macanudo527: Yeah, I've been holding off until I can make sure it works, haha. I **think** this will give us better pricing, once I pull it off. When everything gets ironed out, I'll try to write a coherent comment or add it to the docs.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -493,7 +610,97 @@ def _add_exchange_to_memcache(self, exchange: str) -> None:
...
         self.__exchange_markets[exchange] = current_markets
-        self.__exchange_graphs[exchange] = current_graph
+
+        return current_graph
+
+    # Isolated to be mocked
+    def _get_request_delay(self, exchange: str) -> float:
+        return _REQUEST_DELAYDICT.get(exchange, 0)
+
+    def _optimize_assets_for_exchange(
```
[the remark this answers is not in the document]
eprbell: This function is critical to the algorithm and fairly complex: could you add comments describing the behavior of its main parts?

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -465,7 +550,39 @@ def _add_alternative_markets(self, graph: MappedGraph[str], current_markets: Dic
...
+        optimizations = self._optimize_assets_for_exchange(
+            unoptimized_graph,
+            self.__manifest.first_transaction_datetime,
+            self.__manifest.assets,
+            exchange,
+        )
+        self.__logger.debug("Optimizations created for graph: %s", optimizations)
+
+        exchange_tree: AVLTree[datetime, MappedGraph[str]] = AVLTree[datetime, MappedGraph[str]]()
+        for timestamp, optimization in optimizations.items():
```
eprbell, line 576: My understanding of this loop is that there can be only one optimization per graph at a given timestamp. Is this correct/intended? Can there be a case in which a graph needs two or more optimizations at a given timestamp?

tests/test_plugin_ccxt.py
```diff
@@ -103,28 +128,78 @@
...
         eth.add_neighbor(usdt, 1.0)
+        usdc.add_neighbor(usd, 2.0)
         usdt.add_neighbor(usd, 1.0)
-        usd.add_neighbor(jpy, 2.0)
+        usd.add_neighbor(jpy, 50.0)
+
+        return [beth, btc, eth, gbp, jpy, usdc, usdt, usd]
+
+    @pytest.fixture
+    def test_graph_optimized(self, test_vertex_list: List[Vertex[str]]) -> MappedGraph[str]:
```
eprbell, line 186: Fixtures should not start with "_test"

eprbell (changes requested):

macanudo527:
I just realized `manifest_builder.py` uses multithreading, but I don't think it will actually increase the speed of anything, will it? It needs to be multi-process, right?

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -341,19 +363,43 @@ def get_historic_bar_from_native_source(self, timestamp: datetime, from_asset: s
...
             self.__logger.debug("Retrieved cache for %s/%s->%s for %s", timestamp, from_asset, to_asset, exchange)
             return historical_bar
 
+        historical_bars: Optional[List[HistoricalBar]] = self.find_historical_bars(from_asset, to_asset, timestamp, exchange)
+
+        if historical_bars:
+            returned_bar: HistoricalBar = historical_bars[0]
+            if (timestamp - returned_bar.timestamp).total_seconds() > _TIME_GRANULARITY_STRING_TO_SECONDS[_ONE_DAY]:
+                raise RP2ValueError(
+                    "INTERNAL ERROR: The time difference between the requested and returned timestamps is "
```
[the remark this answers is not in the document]
macanudo527: Yeah, I noticed this toward the end. I'll find/replace.

macanudo527:
This takes an incredibly long time (20-30minutes) to optimize the Kraken graph mostly because of all the fiat pairs. For example, Kraken has markets for BTC/USD, BTC/EUR, BTC/JPY, BTC/GBP all of which need to be optimized. If we only looked at  markets with the quote asset of USD we would cut the time down significantly. If users want, they can optionally add another quote asset.

Another option is to implement multiprocessing since my CPUs are hardly being used during the download and chunk process.

eprbell:
> I just realized `manifest_builder.py` uses multithreading, but I don't think it will actually increase the speed of anything, will it? It needs to be multi-process, right?

In Python CPU-bound algorithms can be sped up using multiprocessing, whereas I/O-bound ones can be sped up with multithreading. This is due to a design issue in the python interpreter (GIL lock). Is this algorithm mostly CPU or I/O-bound?

eprbell:
> 

A couple of ideas:
* could we do internal price lookup in USD only and then convert from USD to the native fiat of the country plugin at the end?
* we could use the profiler to understand where the bottlenecks of the algorithm are and then come up with specific optimizations.

macanudo527:
I'm running into another issue. It looks like Huobi changed something about their API and it no longer seems to work like it should. They do have CSV files available though, so we will eventually be able to price things using that exchange just not now. 

Should I just monkey patch it for now with zero prices for the two assets we need Huobi for? (SOLO and NEXO) In other words, have dali-rp2 insert zero prices with a `info` to logger informing users that the prices aren't available? That way people can at least process some taxes.

Otherwise, there is no other exchange that can price those assets (Binance has the most volume, but might be off-limits).

eprbell:
> I'm running into another issue. It looks like Huobi changed something about their API and it no longer seems to work like it should. They do have CSV files available though, so we will eventually be able to price things using that exchange just not now.
> 
> Should I just monkey patch it for now with zero prices for the two assets we need Huobi for? (SOLO and NEXO) In other words, have dali-rp2 insert zero prices with a `info` to logger informing users that the prices aren't available? That way people can at least process some taxes.
> 
> Otherwise, there is no other exchange that can price those assets (Binance has the most volume, but might be off-limits).

I'd rather not process them than process them with incorrect (zero) price. People who need those two assets priced can still use the manual plug in, until we implement the Huobi CSV plugin. But maybe open an issue so we can track this. Does this make sense?

macanudo527:
Well, I can't test with real world data since those assets get pulled in when I do a pull from Binance.com.

I could sub in the Binance.com route for now maybe. And then try to put together a Huobi plugin soonish.

macanudo527:
Looking around, it looks like Gate.io will do in a pinch for both assets. The volume is only 1% of the total but the price is fairly close and will work until we get something better.

macanudo527:
> > Looking around, it looks like Gate.io will do in a pinch for both assets. The volume is only 1% of the total but the price is fairly close and will work until we get something better.
> 
> Sounds good! Thanks for finding this.

I spoke too soon. The new system doesn't allow the lookup of prices on exchanges where the market doesn't exist, yet. The side effect is that Binance, Gate.io, and even Bitget all added NEXO after I started trading it. Huobi and Bitfinex are the only high-profile exchanges that have it. I'm trying with Bitfinex, now.

macanudo527:
I have another question. I'm wondering how to save the bundle of bars we pull in order to optimize the graph. Normally, we cache individual bars one by one to the cache, but these bars will most likely only be pulled as a bundle, so it makes more sense to store them as a bundle in the cache. 

Should I create a new caching function that takes a different key to store bundles?

The other option is to push it into the Kraken CSV plugin and have it write it out into the CSV. We will always need to pull from the CSV, so we can retrieve all the bars in one call that way. If I store in standard cache to disk, we will need two calls.

eprbell:
> I have another question. I'm wondering how to save the bundle of bars we pull in order to optimize the graph. Normally, we cache individual bars one by one to the cache, but these bars will most likely only be pulled as a bundle, so it makes more sense to store them as a bundle in the cache.
> 
> Should I create a new caching function that takes a different key to store bundles?
> 
> The other option is to push it into the Kraken CSV plugin and have it write it out into the CSV. We will always need to pull from the CSV, so we can retrieve all the bars in one call that way. If I store in standard cache to disk, we will need two calls.

The cache read/write functions already use Any-typed data, so I don't think we need new variants.

macanudo527:
OK, I got the bars caching issue resolved.

About speeding up the optimization process. I've narrowed down the quote assets to what is in the transactions plus the native fiat of the country in the dali configuration. This should narrow it down enough so that optimization only takes around 5 minutes or so for most people and once cached will only take about a minute to load up.

eprbell:
> OK, I got the bars caching issue resolved.
> 
> About speeding up the optimization process. I've narrowed down the quote assets to what is in the transactions plus the native fiat of the country in the dali configuration. This should narrow it down enough so that optimization only takes around 5 minutes or so for most people and once cached will only take about a minute to load up.

Fantastic! We can optionally do a profiler run later to identify further bottlenecks, but for now this is great!

macanudo527:
All right, one more fringe case that the optimizations have uncovered. Some assets are given to users before they are traded. Generally, tax law states that the cost basis is the price of the asset when a market does finally open. However, under the current system, the ccxt pair plugin rejects it because it can't find a price. 

I think I will add a week (the asset I'm having trouble with was given to me 5 days before I could trade it anywhere) to the start of the graph for that asset. Then, when the price is pulled and the timestamps don't match (because the price is from the future) I can use that price but warn the user with an `INFO`.

eprbell:
> All right, one more fringe case that the optimizations have uncovered. Some assets are given to users before they are traded. Generally, tax law states that the cost basis is the price of the asset when a market does finally open. However, under the current system, the ccxt pair plugin rejects it because it can't find a price.
> 
> I think I will add a week (the asset I'm having trouble with was given to me 5 days before I could trade it anywhere) to the start of the graph for that asset. Then, when the price is pulled and the timestamps don't match (because the price is from the future) I can use that price but warn the user with an `INFO`.

Sounds good: you're finding lots of edge cases. Good work!

macanudo527:
I just discovered another issue. The Kraken REST API only returns the latest 720 bars. So, the farther we go back in the current quarter the less accurate the pricing becomes until new quarterly data comes available.  That's probably confusing so let me give an example:

I request a price for BTC/USD for April 2nd, which is in 2Q and not available via CSV yet.CCXT will first request a 1 minute candle and not get it since it only retrieves the last 720 minutes. It continues to increase the timeframe until it can get the candle that includes April 2nd. This is probably 12 hour, since April 2nd is easily within 360 days from now. It will use that candle if I pull the data now, but later when CSV data comes available, it will use the 1 minute candle increasing the accuracy.

a 12 hour candle isn't too bad, but dali-rp2 has to waste calls for each timeframe. Each call takes 5 seconds. That means it can take 20-25 seconds for each transaction to resolve.

I hope that all makes sense. Talking to the Kraken team, they told me that I can construct OHLC from their trades endpoint which will give me all trades and their execution prices during a time period without restrictions, but that seems a little overkill, and maybe not worth it for just one exchange.

eprbell:
I'm back from vacation and I'll try to catch up with everything over the next few days: thanks for holding down the fort! 

Regarding the Kraken 720 bar issue, it seems to me we don't have a lot of choice: we'll have to use less precise prices until the CSV becomes available. Perhaps we can use an info log to let users know that prices from the last 3 months will probably change later. This is not great because it breaks user's golden-file-based test (if they have any), but I'm not sure what else we can do. Can you elaborate on OHLC construction from their trades endpoint? I'd just like to understand how that would work.

macanudo527:
Most exchanges will give you their raw trade log (ie. all trades performed on the exchange and at what price). Here is an example from Kraken:
```
{"error":[],"result":{"XXBTZUSD":[["8552.90000","0.03190270",1559347203.7998,"s","m",""],["8552.90000","0.03155529",1559347203.8086,"s","m",""],["8552.90000","0.00510797",1559347203.9664,"s","m",""],["8552.90000","0.09047336",1559347203.9789,"s","m",""],["8552.90000","0.00328738",1559347203.9847,"s","m",""],["8552.90000","0.00492152",1559347203.9897,"s","m",""],["8552.90000","0.00201848",1559347203.9937,"s","m",""],["8552.90000","0.11422068",1559347203.9993,"s","m",""],["8552.90000","0.00425858",1559347204.071,"s","m",""],["8552.90000","0.00427679",1559347204.0762,"s","m",""],["8552.90000","0.06381401",1559347204.1662,"s","m",""]...["8579.50000","0.05379597",1559350785.248,"s","l",""],["8579.50000","0.94620403",1559350785.2936,"s","l",""],["8578.10000","0.45529068",1559350785.297,"s","l",""]],"last":"1559350785297011117"}}
```

You can then reconstruct the candle by grabbing the open, close, high, low values from the trades in a given window and then add volume. Of course, if you wanted super accurate prices you can literally take the exact trade you made and grab the price from that.

Here is more details from [Kraken on how to pull trades](https://support.kraken.com/hc/en-us/articles/218198197-How-to-retrieve-historical-time-and-sales-trading-history-using-the-REST-API-Trades-endpoint-).

macanudo527:
I implemented the warning, it fills up the console pretty quickly if someone has done a lot of trades, but maybe in an update we can put together some kind of summary.

eprbell:
Thanks for the explanation. Regarding the warding filling up the console: perhaps we can print it only once so that it's easier to read (without transaction specifics, just a message saying that prices from Karaken from the last 3 months are not fully accurate and will improve in the next 3 months). I also suggest adding a FAQ explaining the issue.

Let me know when new code is up and I'll do one more round of review.

src/dali/plugin/pair_converter/ccxt.py
macanudo527: I'm not sure if this was answered earlier, but yes there can be multiple optimizations at a given timestamp. I've done a little more optimization to try to reduce the number of snapshots made. Each snapshot will have all the optimizations of the previous snapshot, but then be updated with any changes, subtractions or additions.

macanudo527:
There seems to be [a bug in the macos build affecting bz2](https://github.com/actions/setup-python/issues/682). Should I temporarily pin python version to 3.7.16?

macanudo527:
Other than the macos bug, this **should** be good to go. I can't fully test it because it just happens Binance.com has a launchpad now for MAV, which means I'm receiving an asset that doesn't have a market and doesn't exist on the graph.

How should I handle that case? Just warn users and assign a zero price, since that is actually its current value. The problem is there is no way of knowing if it is an untradeable asset or we just haven't added an alternative market to price it yet.

macanudo527:
@eprbell I think I'll just allow users to declare untradeable assets in the plugin configuration file. Then, we can assign ZERO to them until a market exists. I'll warn the user with an info.

macanudo527:
This works with real world data, and provides much more accurate pricing.

eprbell:
> This works with real world data, and provides much more accurate pricing.

Great! I'll do one more full review.

macanudo527:
@eprbell It looks like the macos bug has been fixed. You just need to [add a snippet](https://github.com/actions/setup-python/issues/682#issuecomment-1604319803) to the CI.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -423,24 +531,92 @@ def find_historical_bar(self, from_asset: str, to_asset: str, timestamp: datetim
                     self.__logger.debug("Server not available. Making attempt #%s of 10 after a ten second delay. Exception - %s", request_count, exc_na)
                     sleep(10)
 
+            if len(historical_data) > 0:
+                returned_timestamp = datetime.fromtimestamp(int(historical_data[0][0]) / _MS_IN_SECOND, timezone.utc)
+                if (returned_timestamp - timestamp).total_seconds() > _TIME_GRANULARITY_STRING_TO_SECONDS[timeframe] and not all_bars:
+                    if retry_count == len(_TIME_GRANULARITY_DICT.get(exchange, _TIME_GRANULARITY)) - 1:  # If this is the last try
+                        # raise RP2ValueError(f"Internal error: Requested a spot price for {timestamp} but
```
eprbell, line 538: Remove?

eprbell (commented):
I started reviewing your latest changes, but unfortunately Github's "Show changes since your last review" doesn't work for me anymore (it used to). So I have to do a full review from scratch. I intend to finish this by tomorrow (Sunday) night PST.

eprbell:
> I started reviewing your latest changes, but unfortunately Github's "Show changes since your last review" doesn't work for me anymore (it used to). So I have to do a full review from scratch. I intend to finish this by tomorrow (Sunday) night PST.

When I try to see changes since last review I now get: "We went looking everywhere, but couldn’t find those commits".

macanudo527:
It might be because I force pushed a few times. I made a lot of changes from the original plan because some things just didn't work out.

eprbell:
> It might be because I force pushed a few times. I made a lot of changes from the original plan because some things just didn't work out.

Oh, I see. In the future it would be best not to use `push --force`, unless you are in critical situations (e.g. if you're overwriting sensitive data that was checked in by mistake, etc.): it makes reviews considerably harder (especially big ones).

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -230,8 +253,8 @@ def exchange_markets(self) -> Dict[str, Dict[str, List[str]]]:
         return self.__exchange_markets
 
     @property
-    def exchange_graphs(self) -> Dict[str, MappedGraph[str]]:
-        return self.__exchange_graphs
+    def exchange_datetime_graph_tree_dict(self) -> Dict[str, AVLTree[datetime, MappedGraph[str]]]:
```
eprbell, line 256: Could you rename the function something like `get_graph_tree_from_exchange` or similar?

eprbell (changes requested):
Just a nit or two. LGTM!

src/dali/plugin/pair_converter/ccxt.py
macanudo527: How about `exchange_2_graph_tree`? That would be in line with the other properties. 

It only seems to be used for testing actually, it's not really necessary.

eprbell: Sure, sounds good.

macanudo527:
OK, this is clean.

eprbell (approved):
LGTM! Congrats on finishing this big improvement!

PR #147, opened 2023-05-02
Set ODS sheet size as function of number of rows in it

qwhelan:
`dali` currently outputs a large fraction of empty rows, which `rp2` then processes individually and potentially outputs a log message per empty line. Assuming `n` transactions per asset and `k` assets, the current `dali` output is `k * n + 40` rows per asset/sheet and a total of `k^2 * n + 40k` rows per document. Thus, only `1/k` rows are populated and `rp2` runtime explodes while doing useless work.

The above actually is the best case scenario, as it is even worse when the distribution of transactions per asset is unevenly distributed.

I have included a test case but it is a bit of a no-op as I did not want to touch the golden datasets and the existing code passes without this change. Please let me know if you have any suggestions on the preferred testing approach in this case.

In order to make `mypy` happy, I also updated the `ezodf` stubs.

### DIFF
--- a/src/stubs/ezodf/__init__.pyi
+++ b/src/stubs/ezodf/__init__.pyi
@@ -1,0 +1,18 @@
+# Copyright 2023 Christopher Whelan
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .document import newdoc, opendoc  # noqa
+from .table import Table
+
+__all__ = ["opendoc", "newdoc", "Table"]
--- a/src/stubs/ezodf/sheets.pyi
+++ b/src/stubs/ezodf/sheets.pyi
@@ -1,4 +1,4 @@
-# Copyright 2021 eprbell
+# Copyright 2023 Christopher Whelan
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


eprbell (changes requested):
This is a really nice fix: thanks for finding the bug and proposing a PR! 

All CI tests are already passing with this diff (see results at the bottom of the PR). I also tried it with my own personal tax data tests and it works well with those too. There is no need to modify the golden files, because they are diffed on non-empty lines, so this change doesn't affect them.

I have a couple of small nits on the code, but beyond that it's almost ready to merge.

BTW, [RP2 itself would benefit from a similar fix](https://github.com/eprbell/rp2/blob/78e3abc754f0683ed00ad42e83d280b97d8ad9a4/src/rp2/plugin/report/rp2_full_report.py#L508) too, in case you're interested. Thanks again!

qwhelan:
@eprbell I implemented your suggested changes and CI is passing.

I'll probably take a stab at the `rp2` version of this change but have some unsubmitted `dali` changes I'd like to focus on first.

src/stubs/ezodf/__init__.pyi
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 eprbell
```
eprbell, line 1: Update copyright notice to be 2023 and use your name, since this is a new file you introduced.

src/stubs/ezodf/sheets.pyi
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 eprbell
```
eprbell, line 1: Same thing with copyright notice.

src/stubs/ezodf/table.pyi
```diff
@@ -0,0 +1,19 @@
+# Copyright 2021 eprbell
```
eprbell, line 1: Same.

eprbell (changes requested):
There are a couple of copyright notices that need updating, otherwise this LGTM.

qwhelan:
@eprbell Updated the files you flagged but kept `src/stubs/ezodf/document.pyi`'s header the same as git is treating it as a rename rather than a new file. Let me know if that one needs fixing.

eprbell (approved):
Awesome!

PR #148, opened 2023-05-02
Fix typos in pair_converter plugin names

qwhelan:
The documentation for `pair_converter` plugins incorrectly points to the general `ccxt.py` case, not the specialized `ccxt_binance.py` or `ccxt_kraken.py` ones.


macanudo527 (approved):
Thanks for fixing these copy/paste errors! You are cleaning up my mess, sorry about that.

PR #152, opened 2023-05-07
Cache signature parameters during transaction initialization

qwhelan:
`dali` currently invokes `inspect.signature()` on every transaction initialization, which is the majority of the runtime after including https://github.com/eprbell/dali-rp2/pull/151. We start with a runtime of:
```
%timeit InTransaction('test', 'a', 'b', '2023-03-30 00:00:00Z', 'c', 'd', 'e', 'BUY', '1', '1')
97.2 µs ± 4.3 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
And by adding a class-level cache, we cut the runtime by ~2/3:
```
%timeit InTransaction('test', 'a', 'b', '2023-03-30 00:00:00Z', 'c', 'd', 'e', 'BUY', '1', '1')
35.1 µs ± 465 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

### DIFF
--- a/src/dali/abstract_transaction.py
+++ b/src/dali/abstract_transaction.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 from datetime import datetime
-from inspect import signature
-from typing import Callable, Dict, List, NamedTuple, Optional, Union
+from inspect import Parameter, signature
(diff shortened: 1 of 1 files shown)


eprbell (approved):
LGTM (see also discussion at #151). Thanks for submitting PRs!

PR #153, opened 2023-05-07
Provide progress bar when resolving transactions

qwhelan:
Use `progressbar2` to show `dali`'s progress during transaction resolution, which can be particularly slow when fetching prices from the web.

Screenshot from 2023-05-06 17-56-36

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,3 +1,3 @@
 [settings]
-known_third_party = dateutil, ezodf, jsonschema, pytz, python-dateutil, requests, rp2, setuptools, yfinance
+known_third_party = dateutil, ezodf, jsonschema, progressbar2, pytz, python-dateutil, requests, rp2, setuptools, yfinance
 profile = black
--- a/src/dali/transaction_resolver.py
+++ b/src/dali/transaction_resolver.py
@@ -16,6 +16,8 @@
 from datetime import datetime
 from typing import Any, Dict, List, NamedTuple, Optional, cast
 
+from progressbar import ProgressBar
+from progressbar.widgets import AdaptiveETA, Bar, Percentage, Timer
 from rp2.rp2_decimal import ZERO, RP2Decimal
 from rp2.rp2_error import RP2RuntimeError, RP2TypeError, RP2ValueError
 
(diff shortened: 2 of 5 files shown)


src/dali/transaction_resolver.py
```diff
@@ -252,49 +253,52 @@ def resolve_transactions(
...
                 unique_id_2_transactions[AssetAndUniqueId(transaction.asset, transaction.unique_id)] = transaction_list
 
-        for transaction_list in unique_id_2_transactions.values():
-            if len(transaction_list) > 2:
-                raise RP2RuntimeError(
-                    f"Internal error: Attempting to resolve more than two transactions with same {Keyword.UNIQUE_ID.value}: {transaction_list}"
-                )
-            if len(transaction_list) == 1:
-                transaction = _apply_transaction_hint(transaction_list[0], global_configuration)
+        with ProgressBar(max_value=len(transactions)) as progress_bar:
```
eprbell, line 256: Shouldn't this be:
```suggestion
        with ProgressBar(max_value=len(unique_id_2_transactions)) as progress_bar:
```

eprbell (commented):
This is a cool change, however in my case (with a few thousand transactions) the bar advances almost instantaneously: how is it for you? Maybe add the first for loop (at line 235 inside the `with ProgressBar`).

qwhelan:
@eprbell A few thousand shouldn't be instantaneous unless you've got all prices in cache, network queries disabled, or all prices populated in your inputs. For a rate limited pricing API, that should be at most a few hundred seconds of run time, so I'm guessing one of the above applies. For myself, it's definitely hours.

macanudo527:
@qwhelan this is huge. This was on my "I'll get to it someday" list, so it is good to see somebody get in here and get it done. I can say from experience that with the Kraken API it takes a good couple of hours and this will really help. 

Can plugins still output debugger info messages? I know the CSV downloader for Kraken will output a message when it downloads and parses a CSV file. Does that interrupt the progress bar?

qwhelan:
@macanudo527 Yeah, they can but that effectively triggers a snapshot of the progress bar to be output, the debug message to be logged, and then the progress bar resumes. So not the prettiest output, but there's settings in `progressbar2` that might let that behavior be refined a bit. The ETA also swings around quite a bit as it uses a rolling 10 event window to forecast that I'd probably want to expand to get more stable ETAs.

eprbell:
> @eprbell A few thousand shouldn't be instantaneous unless you've got all prices in cache, network queries disabled, or all prices populated in your inputs. For a rate limited pricing API, that should be at most a few hundred seconds of run time, so I'm guessing one of the above applies. For myself, it's definitely hours.

Duh, it was the cache :-)

eprbell:
> 

@macanudo527, can you try it out with your setup and see how the output looks? In my case it's fine, but I just use the historic crypto plugin which doesn't print anything during transaction resolution.

qwhelan:
@eprbell Just fixed the typo you flagged and added a second progress bar to the first for loop that you suggested.

src/dali/transaction_resolver.py
```diff
@@ -232,69 +233,73 @@ def resolve_transactions(
     transaction: AbstractTransaction
 
     try:
-        for transaction in transactions:
-            if not isinstance(transaction, AbstractTransaction):
-                raise RP2RuntimeError(f"Internal error: Parameter 'transaction' is not a subclass of AbstractTransaction. {transaction}")
+        with ProgressBar(max_value=len(transactions)) as progress_bar:
+            for transaction in transactions:
```
eprbell, line 237: I think this loop needs to call `progress_bar.update()` as well.

src/dali/transaction_resolver.py
```diff
@@ -232,69 +233,73 @@ def resolve_transactions(
...
+                        transaction = _update_spot_price_from_web(transaction, global_configuration)
+                    LOGGER.debug("Unresolvable transaction (no %s): %s", Keyword.UNIQUE_ID.value, str(transaction))
+                    resolved_transactions.append(transaction)
+                else:
+                    transaction_list: List[AbstractTransaction]
+                    transaction_list = unique_id_2_transactions.setdefault(AssetAndUniqueId(transaction.asset, transaction.unique_id), [])
+                    transaction_list.append(transaction)
+                    unique_id_2_transactions[AssetAndUniqueId(transaction.asset, transaction.unique_id)] = transaction_list
+
+        with ProgressBar(max_value=len(unique_id_2_transactions.values())) as progress_bar:
```
eprbell, line 257: Would it be better to have just one progress bar? In other words have only one `with ProgressBar` at line 236, but initialize it this way:
```
        with ProgressBar(max_value=len(transactions) + len(unique_id_2_transactions)) as progress_bar:
```
And place both loops in the scope of the `with`. 

Also, would it be possible to display just the percentage, without the absolute number (which may or may not make sense to the user)?

eprbell (changes requested):

src/dali/transaction_resolver.py
qwhelan: Yes, that'd be better. The issue is that `unique_id_2_transactions` isn't populated until the first for loop, but I think it might be possible to set `max_value=2*len(transactions)` and then update `max_value` once the first for loop is done (so might jump from 50% to 100% if the latter is empty).

And yes, the output is totally configurable but I might not be able to finalize that today, so it might be a few days before this is ready to merge.

macanudo527: The two progress bars are a bit confusing. I think updating the max_value midway sounds like a reasonable solution. We don't need to get too fancy.

macanudo527:
This works well enough with my setup. My only issue is that the Info messages get tacked on to the end of the progress bar. They don't get displayed on their own line. I could add a new line to every info message, but that isn't ideal of course. 

Or, it seems like `progressbar2` has a better solution for this since it wraps the stderr.

It would be more useful if the adaptiveETA sampled the last 100 transactions since the ETA jumps around a lot. The regular ETA may actually work better for us.

I'm not too concerned about having a very pretty output, though.

qwhelan:
@macanudo527 I can't get any messages to show up with my setup after adding the `wrap_stderr()` logic, could you check that you're seeing the expected behavior now?

I've updated the widgets to remove the raw counts and switch the `AdaptiveETA` to use 100 data points. Using the raw `ETA` caused some issues if the first `for` loop is very fast relative to the second (which is true for my data) as that causes the ETA to be massively biased low.
Screenshot from 2023-05-21 22-01-12

src/dali/transaction_resolver.py
```diff
@@ -16,6 +16,7 @@
 from datetime import datetime
 from typing import Any, Dict, List, NamedTuple, Optional, cast
 
+from progressbar import ProgressBar, widgets
```
eprbell, line 19: Rather than importing the entire widget module, could you import the single classes that you need (Percentage, Bar, etc.)?

src/dali/transaction_resolver.py
```diff
@@ -231,70 +232,90 @@ def resolve_transactions(
     unique_id_2_transactions: Dict[AssetAndUniqueId, List[AbstractTransaction]] = {}
     transaction: AbstractTransaction
 
-    try:
-        for transaction in transactions:
-            if not isinstance(transaction, AbstractTransaction):
-                raise RP2RuntimeError(f"Internal error: Parameter 'transaction' is not a subclass of AbstractTransaction. {transaction}")
+    progress_bar = ProgressBar(
+        max_value=2 * len(transactions),
```
eprbell, line 236: Add space before and after `=` (here and below).

Also why is `max_value` twice the length of transactions? Is it to approximate `len(transactions) + len(unique_id_2_transactions)`? If so: clever!

src/dali/transaction_resolver.py
```diff
@@ -231,70 +232,90 @@ def resolve_transactions(
...
-                transaction = _resolve_out_in_transaction(transaction1, transaction2, None)
-            elif isinstance(transaction1, IntraTransaction) and isinstance(transaction2, IntraTransaction):
-                transaction = _resolve_intra_intra_transaction(transaction1, transaction2, None)
-            else:
-                raise RP2RuntimeError(
-                    f"Internal error: attempting to resolve two transactions that aren't Intra/Intra, In/Out or Out/In:\n{transaction1}\n{transaction2}"
-                )
-
-            if read_spot_price_from_web:
-                transaction = _update_spot_price_from_web(transaction, global_configuration)
```
eprbell, line 294: These two lines should not be removed.

eprbell (changes requested):
A few nits and one fix needed, but it looks pretty good to me.

.isort.cfg
```diff
@@ -1,3 +1,3 @@
 [settings]
-known_third_party = dateutil, ezodf, jsonschema, pytz, python-dateutil, requests, rp2, setuptools, yfinance
+known_third_party = dateutil, ezodf, jsonschema, progressbar, pytz, python-dateutil, requests, rp2, setuptools, yfinance
```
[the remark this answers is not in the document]
qwhelan: This is actually unnecessary but was assuming the import-able name was the relevant one for sorting purposes (which is `progressbar`)

src/dali/transaction_resolver.py
qwhelan: `black` doesn't like spaces here. I could define the arguments a few lines above, which would get spaces.

And yes - the idea was to over-estimate so the user only gets positive surprises in progress/ETA.

src/dali/transaction_resolver.py
qwhelan: I think this is showing up as removed due to indentation changes, viewing with whitespace off shows this being kept.

src/dali/logger.py
```diff
@@ -15,6 +15,10 @@
 
 import logging
 
+import progressbar
```
eprbell, line 18: Can you import directly the symbol you're going to use (wrap_stderr)?

eprbell (changes requested):
Only one nit. If it works with @macanudo527's setup, we can merge.

src/dali/logger.py
qwhelan: No, actually - `progressbar.streams` is an object rather than a module: https://github.com/wolph/python-progressbar/blob/develop/progressbar/utils.py#L531

I've made the import as granular as feasible in the most recent push.

macanudo527:
Sorry for the late reply. This looks great to me!

Issue #157, opened 2023-05-11
2 same uniqueID transactions in binance´s dust trade.


Larision:
I was checking historical transactions and i can confirm was a dust trade.
Checking all dust trades only BTC, USDT, BUSD and EUR have pair assets like BNB/XXX, all others are XXX/BNB and getting first asset to append to id number like id=1111111111BTC.
Here i traded usdt and btc dust in bnb so both got BNB append making them have the same uniqueID making dali failing resolving them..
Srry if im a bit lost but im quite newbie with python.

```
Traceback (most recent call last):
  File "/home/kali/renta/dali-rp2/src/dali/dali_main.py", line 168, in _dali_main_internal
    resolved_transactions: List[AbstractTransaction] = resolve_transactions(transactions, dali_configuration, args.read_spot_price_from_web)
                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kali/renta/dali-rp2/src/dali/transaction_resolver.py", line 289, in resolve_transactions
    raise RP2RuntimeError(
rp2.rp2_error.RP2RuntimeError: Internal error: attempting to resolve two transactions that aren't Intra/Intra, In/Out or Out/In:
InTransaction:
  plugin=Binance.com_REST
  unique_id=XXX760BNB
  raw_data={"id": "XXX760BNB", "timestamp": 1648564129000, "datetime": "2022-03-29T14:28:49.000Z", "symbol": "BNB/USDT", "order": "XXX760", "type": null, "takerOrMaker": null, "side": "buy", "amount": 8.313e-05, "price": 439.07133405509444, "cost": 0.0365, "fee": {"currency": "BNB", "cost": 1.66e-06}, "info": {"fromAsset": "USDT", "amount": "0.0365", "transferedAmount": "0.00008313", "serviceChargeAmount": "0.00000166", "operateTime": "1648564129000", "transId": "XXX760", "isDustTrade": true}}
  timestamp=2022-03-29 14:28:49+0000
  asset=BNB
  exchange=Binance.com
  holder=fran
  transaction_type=Buy
  spot_price=__unknown
  crypto_in=0.00008313
  fiat_fee=None
  fiat_in_no_fee=None
  fiat_in_with_fee=None
[...]

PR #160, opened 2023-05-15
Fix Dust Trade ID naming for markets that have BNB as the quote asset

macanudo527:
This addresses #157 . With some assets, Binance.com uses a market with BNB as the quote asset instead of as the base asset. This produced duplicate IDs.


eprbell (approved):
LGTM!

Issue #166, opened 2023-05-18
Bitbank widthdrawals status CONFIRM_TIMEOUT


topherbuckley:
I got the following error while importing a csv from bitbank. 

```
ERROR: Fatal exception occurred:
Traceback (most recent call last):
  File "/media/christopher/HDD/git/dali-rp2/src/dali/dali_main.py", line 176, in _dali_main_internal
    result_list = pool.map(_input_plugin_helper, input_plugin_args_list)
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 367, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 774, in get
    raise self._value
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/media/christopher/HDD/git/dali-rp2/src/dali/dali_main.py", line 214, in _input_plugin_helper
    plugin_transactions = input_plugin.load(country)
  File "/media/christopher/HDD/git/dali-rp2/src/dali/plugin/input/csv/bitbank_supplemental.py", line 73, in load
    result.extend(self.parse_withdrawals_file(self.__withdrawals_csv_file))
  File "/media/christopher/HDD/git/dali-rp2/src/dali/plugin/input/csv/bitbank_supplemental.py", line 140, in parse_withdrawals_file
    IntraTransaction(
  File "/media/christopher/HDD/git/dali-rp2/src/dali/intra_transaction.py", line 50, in __init__
    super().__init__(
  File "/media/christopher/HDD/git/dali-rp2/src/dali/abstract_transaction.py", line 115, in __init__
    self.__unique_id: str = self._validate_string_field(Keyword.UNIQUE_ID.value, unique_id, raw_data, disallow_empty=True, disallow_unknown=False)
  File "/media/christopher/HDD/git/dali-rp2/src/dali/abstract_transaction.py", line 55, in _validate_string_field
    raise RP2RuntimeError(f"Internal error: {name} is empty: {raw_data}")
rp2.rp2_error.RP2RuntimeError: Internal error: unique_id is empty: 2021/09/12 20:10:27,3020,0.01,3020.01,Coinbase,{RECATED_ADDRESS},,CONFIRM_TIMEOUT

```

[...]

PR #167, opened 2023-05-18
Adding if logic to check it the status on bitbank withdrawals is DONE…

topherbuckley:
… or not before proceeding to parse

As per [issue 166](https://github.com/eprbell/dali-rp2/issues/166)


topherbuckley:
Just wanted to note that a similar thing is done for the deposits [here](https://github.com/eprbell/dali-rp2/blob/e8775b12b167e4d4053550ffbba6cead5044caa2/src/dali/plugin/input/csv/bitbank_supplemental.py#L91), and this is what I did in order to keep the handling of these similar.

macanudo527 (approved):
This was obviously just an oversight by me. Thanks for the patch!

Issue #159, opened 2023-05-12
Spot price for XXX->YYY not found on any pair converter plugin


topherbuckley:
# TLDR
Gate.io doesn't support BOBA/USD direct conversion, but it looks like this is manually added to the ccxt.py. |

# All the details:
Getting the following error when running the bitbank rest api

```
ERROR: Fatal exception occurred:
Traceback (most recent call last):
  File "/media/christopher/HDD/git/dali-rp2/src/dali/dali_main.py", line 181, in _dali_main_internal
    resolved_transactions: List[AbstractTransaction] = resolve_transactions(transactions, dali_configuration, args.read_spot_price_from_web)
  File "/media/christopher/HDD/git/dali-rp2/src/dali/transaction_resolver.py", line 263, in resolve_transactions
    transaction = _update_spot_price_from_web(transaction, global_configuration)
  File "/media/christopher/HDD/git/dali-rp2/src/dali/transaction_resolver.py", line 144, in _update_spot_price_from_web
    raise RP2RuntimeError(
rp2.rp2_error.RP2RuntimeError: Spot price for 1224573781:2022-05-10 06:20:39+00:00:BOBA->USD not found on any pair converter plugin
```

For context, the lines around `transaction_resolver.py` (135-148) are as follows:

```python
    if is_unknown(transaction.spot_price) or RP2Decimal(transaction.spot_price) == ZERO:  # type: ignore
        conversion: RateAndPairConverter = _get_pair_conversion_rate(
            timestamp=transaction.timestamp_value,
            from_asset=transaction.asset,
            to_asset=global_configuration[Keyword.NATIVE_FIAT.value],
            exchange=_get_originating_exchange(transaction),
            global_configuration=global_configuration,
        )
        if conversion.rate is None:
            raise RP2RuntimeError(
                f"Spot price for {transaction.unique_id + ':' if transaction.unique_id else ''}"
                f"{transaction.timestamp_value}:{transaction.asset}->{global_configuration[Keyword.NATIVE_FIAT.value]}"
                " not found on any pair converter plugin"
            )
```
Jumping into `_get_pair_conversion_rate`:

```python
def _get_pair_conversion_rate(timestamp: datetime, from_asset: str, to_asset: str, exchange: str, global_configuration: Dict[str, Any]) -> RateAndPairConverter:
    rate: Optional[RP2Decimal] = None
    pair_converter: Optional[AbstractPairConverterPlugin] = None
[...]

PR #171, opened 2023-05-19
Adding upbit exchange via ccxt

topherbuckley:
As per [Issue 159](https://github.com/eprbell/dali-rp2/issues/159#issuecomment-1553862095)

@macanudo527 Seems to have worked. Is this the intended way of doing this though?

### DIFF
--- a/src/dali/plugin/pair_converter/ccxt.py
+++ b/src/dali/plugin/pair_converter/ccxt.py
@@ -30,6 +30,7 @@ from ccxt import (
     gateio,
     huobi,
     kraken,
+    upbit,
 )
 from prezzemolo.vertex import Vertex
 from rp2.logger import create_logger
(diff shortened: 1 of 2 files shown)


macanudo527:
This should work, but did you add the market? like you did for BOBA and XYM?

topherbuckley:
@macanudo527, yeah I added them, and tested, but thought I'd keep those for a separate commit just to be cleaner as upbit support is separate from support of a specific coin, but happy to add that here as well if preferred.

topherbuckley:
I just added the ARK/KRW pair, but let me know if you'd like me to take those back out and add back into a PR containing a bunch of other pairs I'll likely be adding.

topherbuckley:
Sorry, was I supposed to add a subclass for each new exchange like is done for [here](https://github.com/eprbell/dali-rp2/tree/main/src/dali/plugin/pair_converter) for binance, coinbase_pro, and kraken? Happy to do so, but wasn't sure if necessary or not (as code ran without them). I recall reading in another issue that these are for clarity, so happy to add them if needed.

macanudo527 (approved):
We shouldn't need a locked exchange for upbit. We can add one later if someone is interested.

topherbuckley:
I have a bunch more of these locally in their own linear branch history as separate branches. Do you want them in a single PR or each to their own?

PR #172, opened 2023-05-19
Automated reformat from black resulted in this

topherbuckley:
Seems like a lot. Is this normal? I've never used black before, and was just following the unit tests on the dev README (in order) and it kept resulting in these changes, so thought I'd commit them. 

If these aren't desired changes, should I try to learn about black and make any necessary changes to the config of how that is working? Any help in the right direction appreciated!

### DIFF
--- a/src/dali/plugin/input/rest/kraken.py
+++ b/src/dali/plugin/input/rest/kraken.py
@@ -22,26 +22,22 @@
 
 import logging
 from datetime import datetime
-from typing import Dict, List, Optional, Union, Set, Tuple, Any
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


eprbell:
No, it's all good: thanks for running black. I like having periodic black runs to ensure code is consistent style-wise. In this case the changes are mostly in newer files (Kraken), which makes sense, as they probably didn't go through black yet. However before merging, let's get @macanudo527's approval too, because he has a large PR in the pipeline and I don't want to make his life more difficult too many conflicts.

macanudo527:
Yeah, this is all Kraken. It shouldn't affect my PR, I'll just have to add a blank `optimize()` to it.

eprbell (approved):
LGTM

PR #194, opened 2023-06-12
coinbase removed from ccxt.py

Larision:


macanudo527 (approved):
Thanks for this @Larision !

PR #195, opened 2023-07-04
Bump Python version to 3.8

macanudo527:
This is to bump the minimum version of Python from 3.7 to 3.8

### DIFF
--- a/.github/workflows/unix_unit_tests.yml
+++ b/.github/workflows/unix_unit_tests.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.8, 3.9, "3.10", "3.11"]
     env:
       PYTHONPATH: ./src:./test
     steps:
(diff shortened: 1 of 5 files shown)


eprbell (changes requested):
LGTM, but there are a few more occurrences that need the version updated:
* README.dev.md:56:DaLI has been tested on Ubuntu Linux, macOS and Windows 10 but it should work on all systems that have Python version 3.7.0 or greater. Virtualenv is recommended for DaLI development.
* README.dev.md:87:First make sure [Python](https://python.org) 3.7 or greater is installed (in the Python installer window be sure to click on "Add Python to PATH"), then open a PowerShell window and enter the following commands:
* README.dev.md:100:* install python 3.7 or greater
* README.md:71:DaLI has been tested on Ubuntu Linux, macOS and Windows 10 but it should work on all systems that have Python version 3.7.0 or greater.
* README.md:102:First make sure [Python](https://python.org) 3.7 or greater is installed (in the Python installer window be sure to click on "Add Python to PATH"), then open a PowerShell window and enter the following:
* README.md:109:* install python 3.7 or greater

macanudo527:
@eprbell Totally forgot about the manual. It should be good to go now.

eprbell (approved):
LGTM!

Issue #161, opened 2023-05-15
Kraken Rest API _initialize_markets broken


JasonSimms:
I just saw the new merge and wanted to give it a go.  Seems to break on kraken.py 

  File "/home/jason/.local/lib/python3.8/site-packages/dali/plugin/input/rest/kraken.py", line 135, in _initialize_markets
    self.base_id_to_base.update({market[_BASE_ID]: market[_BASE]})
TypeError: list indices must be integers or slices, not str.

Also thanks for the work of this implemenation.  I've gotta get this done this week! Yikes. ----

LOG ---   The log debug print is the market variable

INFO: Reading crypto data using plugin 'dali.plugin.input.rest.kraken'
DEBUG initialize markets:>
debug the market in markets...
[...]

PR #197, opened 2023-08-13
Fix/initialize markets

alexeckert:
Closes https://github.com/eprbell/dali-rp2/issues/161

PR #198, opened 2023-08-26
fixed kraken:_initialize_markets and bug associated with Issue #161;

ndopencode:
added unit tests; original issue was due to ccxt versions differing in the data object for markets_by_id. Conformed kraken ccxt plugin to the latest ccxt version supported by DALI.

Closes https://github.com/eprbell/dali-rp2/issues/161


ndopencode:
Looking at the unit test errors, I'm not sure how to address them since they are outside of the code of the kraken input plugin and its unit test. Locally, all CI checks pass without any failure so I'm not sure why the server CI system is not able to push through the "too many requests" error. 

`requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://api.exchangerate.host/symbols`

macanudo527:
> Looking at the unit test errors, I'm not sure how to address them since they are outside of the code of the kraken input plugin and its unit test. Locally, all CI checks pass without any failure so I'm not sure why the server CI system is not able to push through the "too many requests" error.
> 
> `requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://api.exchangerate.host/symbols`

That test is just flakey. We will re-run it.

mypy.ini
```diff
@@ -30,7 +30,6 @@ pretty = True
 show_column_numbers = True
 show_error_codes = True
 show_error_context = True
-show_none_errors = True
```
eprbell, line 33: This disables the check globally. Can you disable the check locally in a single file instead, if you need to?

src/dali/plugin/input/rest/kraken.py
```diff
@@ -122,10 +122,31 @@ def _initialize_client(self) -> kraken:
...
+                exc_str = f"Expected List from Kraken CCXT Exchange, got {type(markets)} instead. " \
+                          f"Incompatible CCXT library - make sure to follow Dali setup instructions " \
+                          f"to install appropriate versions of dependencies."
+                raise RP2RuntimeError(exc_str)
+
+            for market in markets:  # type: ignore
+                base_id: str = market[_BASE_ID]
+                if base_id in self.base_id_to_base and \
+                    market[_BASE] != self.base_id_to_base[base_id]:
+                    self.__logger.warning(
```
eprbell, line 140: Can you clarify in a comment how this happens? I'm also trying to understand how bad this is: is it just a warning (recoverable) or is it a RP2RuntimeError?

src/dali/plugin/input/rest/kraken.py
```diff
@@ -230,8 +251,17 @@ def _compute_transaction_set(self, trade_history: Dict[str, Dict[str, str]], led
 
             if record[_TYPE] == _TRADE and not is_fiat_asset:
                 self.__logger.debug("Trade history record: %s", trade_history[record[_REFID]])
-                market: Dict[str, str] = self._client.markets_by_id[trade_history[record[_REFID]][_PAIR]]
-                asset_quote: str = market[_QUOTE]
+                markets: List[Dict[str, str]] = self._client.markets_by_id[trade_history[record[_REFID]][_PAIR]]  # type: ignore
+                if len(markets) > 1:
+                    possible_quotes = [market[_QUOTE] for market in markets]
+                    self.__logger.warning(
+                        "Multiple markets for pair. Which quote to use for %s market? Defaulting to quote=%s "
```
eprbell, line 258: Ditto: could you add a comment describing the reasons that can cause this warning?

eprbell (changes requested):
Thanks for submitting this PR! @alexeckert also has an ongoing PR trying to fix this (see #197). @alexeckert, could you try this PR with your data to see if it fixes the issue?

mypy.ini
ndopencode: Oh, I don't need to. I saw this error in my local CI. Maybe it was the wrong to do. 

`mypy.ini: [mypy]: Unrecognized option: show_none_errors = True`

src/dali/plugin/input/rest/kraken.py
ndopencode: I've never seen it happen but because `ccxt` passes back a list I didn't want it to go unnoticed. I'm assuming `ccxt` may have an exchange that has multiple markets for a pair. It's not clear what prompted them to implement it this way but in case kraken did do it I added the warning to log it. Would it make more sense to make it an exception? The other option is to remove it as it may be the result of another exchange.

src/dali/plugin/input/rest/kraken.py
ndopencode: same reason as before. I can add a comment to address but should I keep this defensive code, change it to exception or just add a comment.

eprbell: Defensive code is great: same observation as above.

mypy.ini
eprbell: Ah, gotcha now. You're right: it's good to remove this.

src/dali/plugin/input/rest/kraken.py
ndopencode: I would side with it causing an issue with computation. Incorrectly categorizing the asset would affect the cost basis and taxable event matching. Having said that I'll make sure to raise an exception and add a comment explaining how this could happen.

eprbell: Sounds like a good plan!

ndopencode:
Updated per comments. Hope the CI checks are all green.

macanudo527:
@eprbell This looks good to me. I got the tests to pass.

eprbell (approved):
Thanks for fixing this! It's great to see the Kraken plugin hardening. LGTM.

PR #201, opened 2023-09-17
fix: don't specify gift for a generic transaction type

iloveitaly:


eprbell:
Welcome back! The documentation check is due to a dead link. I disabled the check for it: could you rebase your PRs and upload the code again?

eprbell (commented):
I'm not sure I see the point of this change: Gift is an arbitrary choice, but sell is also an arbitrary choice. The user can control the final type using [transaction hints](https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md#transaction-hints-section).

PR #207, opened 2023-09-25
Fix Pricing Bug when Using Default Exchange

macanudo527:
### What?
When the default Kraken exchange was assigned to another exchange the Kraken CSV reader was not being used. This would cause prices with longer timeframes to be used for prices further in the past.

This PR fixes that bug and prevents double instantiation of the CSV reader.

Also included in this PR are some minor formatting fixes to previously submitted code.

PR #203, opened 2023-09-17
feat: add txn id on nexo transactions

iloveitaly:

PR #212, opened 2023-10-05
Implement Access Keys for Exchangerates.host

### LINKED ISSUE #211: Implement Authentication for Exchangerate.host
### What? Why?

Exchangerate.host, the site that provides fiat prices has recently implemented access keys in order to access their pricing services. Users will be required to provide access keys in order to convert fiat amounts. The free tier allows 1000 requests/month, which should be enough for most users.

Came up in #207 
Will probably fix #209 

### Steps to Reproduce
Try to pull any kind of report in dali_rp2 that requires a fiat conversion e.g. USD->EUR

### Expected Behavior
Conversion of Fiat should be handled by exchangerate.host

### Actual Behavior
Dali-RP2 exits with an error stating that the conversion can not be found on any plugin.

### Priority/Severity
high - Reporting of most trades in a currency other than USD is affected.

### References
[Exchangerate.host](https://exchangerate.host/)
[Exchangerate.host Plans](https://exchangerate.host/product)
[Exchangerate.host API documentation](https://exchangerate.host/documentation)

macanudo527:
This PR will significantly speed up testing and make them more performant. It adds Vcrpy to Dali-RP2, so that we can mock HTTP requests as well as adds support for the now necessary access keys for the fiat pricing provider Exchangerates.host.  

Fixes #211


macanudo527:
@eprbell It took some finagling, but I got vcrpy to work. It does not really like Python < 3.10.  The package itself is amazing. It really speeds things up and is easy to use. It cuts down on a lot of bulky mocking code.

setup.cfg
```diff
@@ -26,6 +26,7 @@ project_urls =
     User Documentation = https://github.com/eprbell/dali-rp2/blob/main/README.md
     Contact = https://eprbell.github.io/eprbell/about.html
 
+# vcpy > 4.4.0 is only compatible with Python 3.10+
```
eprbell, line 29: Do we need to drop support for 3.8 and 3.9 in .github/workflows (I'm OK with it)?

src/dali/abstract_pair_converter_plugin.py
```diff
@@ -58,6 +60,8 @@
 _STANDARD_WEIGHT: float = 1
 _STANDARD_INCREMENT: float = 1
 
+_CONFIG_FILE_URL: str = "https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md"
```
eprbell, line 63: I suggest:
```suggestion
_CONFIG_FILE_DOC_URL: str = "https://github.com/eprbell/dali-rp2/blob/main/docs/configuration_file.md"
```

tests/conftest.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2023 Neal Chambers
```
eprbell, line 1: Could you rename this file to something more descriptive?

tests/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, List, Union
+
+import pytest
+
+
+@pytest.fixture(scope="session")
```
eprbell, line 20: Can you add a comment describing what this file does?

eprbell (commented):
Great! I have just a few nits here and there but otherwise it LGTM.

setup.cfg
macanudo527: I guess 3.10 came out late 2020, so it is probably not so unreasonable to ask users to upgrade. It's not such a big issue now that I worked around it, though.

tests/conftest.py
macanudo527: I'm pretty sure that is what it has to be named in order to run before all tests are run. There isn't a lot of documentation on it, but that is what [this comment](https://stackoverflow.com/questions/34466027/what-is-conftest-py-for-in-pytest#:~:text=Yeah%2C%20the%20documentation,at%2014%3A46) suggests. Pytest documentation just has this [snippet](https://docs.pytest.org/en/7.1.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files) on it for fixtures, but pytest-recording suggests this file as where to put the VCR config.

setup.cfg
eprbell: I was confused by the comment, which suggests that vrcpy needs Python 3.10 +, so I thought we had to drop support for older versions. Is this correct?

setup.cfg
```diff
@@ -39,10 +40,12 @@ install_requires =
     prezzemolo>=0.0.4
     progressbar2>=4.2.0
     pyexcel-ezodf>=0.3.4
+    pytest-recording==0.13.0
```
eprbell, line 43: Use >=

setup.cfg
```diff
@@ -39,10 +40,12 @@ install_requires =
...
     python-dateutil>=2.8.2
     pytz>=2021.3
     requests>=2.26.0
     rp2>=1.5.0
+    vcrpy==4.4.0
```
eprbell, line 48: Use >=

src/dali/abstract_pair_converter_plugin.py
eprbell: The identifier is still the old one: unresolving the conversation.

tests/cassettes/test_plugin_ccxt/unified.yaml
```diff
@@ -0,0 +1,98 @@
+interactions:
```
eprbell, line 1: Could you rename this file to something more descriptive and add a comment?

tests/conftest.py
eprbell: Unresolving: the comment is not there yet.

eprbell (changes requested):
A few comments here and there: getting there!

setup.cfg
macanudo527: Right now, vcr 5.0+ is available, but if we keep it locked to 4.4.0 we can still use Python 3.8, 3.9. If I unlock it, we have to drop support for Python 3.8/9.  That's why the `setup.cfg` file has == and not >=. 

Basically, I really don't need it to be 5.0. It doesn't add that much and this is not mission critical, so I don't need Python 3.10. It works as is. However, if you feel like it is time to drop support I can set it to use the newer versions.  I don't feel strongly either way really. We might have issues with people having problems with not having the right Python version, though.

Anyway, I was waiting for your answer to drop or not to drop before submitting the other changes that is why they are not reflected, yet.

tests/conftest.py
macanudo527: This is resolved locally, just waiting for your response to the python question before I commit.

src/dali/abstract_pair_converter_plugin.py
macanudo527: This is resolved locally, just waiting for your response to the python question before I commit.

setup.cfg
eprbell: Ah, OK, I had misinterpreted the comment. It's OK to fix the version to 4.4 if it allows us to use older Python versions. Later we'll drop 3.8 and 3.9 but it's not urgent right now.

macanudo527:
@eprbell Sorry about the confusion. I should have made a comment instead of pushed the request review button. I've been snowed over with this new job at times. I was in a rush.

Hopefully, this is all good to go now. Let me know if you need anything else. After this gets rolled through, I'll try to rebase the pricing correction #207 and then if we rebase #203 it should also go green.

eprbell (approved):
LGTM! No worries, I know what it's like to start a new coding job (I hope that it's going well though). Thanks for fixing this!

Issue #200, opened 2023-09-12
Implement BitGet REST API Data Loader CCXT Plugin 


macanudo527:
Writing a DaLI data loader plugin has high impact on the functionality and usefulness of both [DaLI](https://github.com/eprbell/dali-rp2) and [RP2](https://github.com/eprbell/rp2). Data loader plugins are well-defined, encapsulated modules that translate native exchange REST API-based data into DaLI's standard format. As such they are good first issues for newcomers to the the project.

Before starting, please read the [contributing guidelines](https://github.com/eprbell/dali-rp2/blob/main/CONTRIBUTING.md#contributing-to-the-repository).

Plugin development is described in the [developer documentation](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md), in particular read the [Dali Internals](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#internal-design) and [Plugin Development](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#plugin-development) sections, which contain all the information needed to build a new data loader plugin.

BitGet REST API is documented here: https://bitgetlimited.github.io/apidoc/en/spot/#welcome

The [Binance plugin](https://github.com/eprbell/dali-rp2/blob/main/src/dali/plugin/input/rest/binance_com.py) can be used as an example.

Before submitting a PR for the new plugin, make sure to go through the [DaLI Plugin Laundry List](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#plugin-laundry-list).

macanudo527:
This just got a lot easier. I found out BitGet has an API method for taxes that joins and pulls all relevant tax records. An exchange that actually makes it easier to do taxes! Go figure.

eprbell:
Oh, interesting: what's the format like? BTW, unrelated, but thanks for going through all and trying to close all the old bugs that got fixed with your latest commits!

macanudo527:
It just sorts all the transactions with a fee, amount and ticker. It then labels them appropriately, buys/sells, etc... Here's the link:

https://bitgetlimited.github.io/apidoc/en/spot/#spot-account-record

I haven't dove into the raw data yet, but looks pretty simple. It might be tough to match up buys and sells though.

macanudo527:
In other news, CCXT has implemented [an auto-pagination feature](https://github.com/ccxt/ccxt/wiki/Manual#automatic-pagination). Currently experimental, but will come in handy maybe next year-ish.

PR #213, opened 2023-11-07
Add Id Pagination

macanudo527:
## What? Why?

Part of #200. 

Id pagination must be implemented to pull data from BitGet. Id pagination is also required for other exchanges. This includes implementation as well as tests.

FYI
Since I'm quite busy these days. I'll be implementing the BitGet issue piecemeal.


eprbell (approved):
Sounds good: piecemeal implementations are better than monolithic ones because they're easier to review.

PR #214, opened 2023-11-08
Add Github Templates

macanudo527:
## What?

I just found these extremely useful in the other projects I've been working on, so I modified a few so they apply to our project and help keep things sorted and moving forward.

I'm open to suggestions if you think something is missing or if they are too much.


eprbell:
Thanks for adding this. I haven't used these templates before, but let's activate them and see how it goes.

eprbell (approved):

PR #221, opened 2024-01-27
Update README.md: Installation on Windows 10 and Other Unix-like Systems

jayr0d:
For Windows 10, Microsoft Visual C++ 14.0 or greater is needed for multidict.

For Other Unix-like Systems, made formatting consistent with other sections.


jayr0d:
Will submit another that includes updates to README.dev.md

PR #222, opened 2024-01-27
Update README*.md: Windows 10 and Other Unix-like Systems

jayr0d:
For Windows 10, Microsoft Visual C++ 14.0 or greater is needed for multidict.  Account for running scripts being disabled.

For Other Unix-like Systems, made formatting consistent with other sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,11 @@ pip install dali-rp2
 ```
 
 ### Installation on Windows 10
-First make sure [Python](https://python.org) 3.8 or greater is installed (in the Python installer window be sure to click on "Add Python to PATH"), then open a PowerShell window and enter the following:
+First make sure [Python](https://python.org) 3.8 or greater is installed (in the Python installer window be sure to click on "Add Python to PATH").
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -108,7 +112,12 @@ pip install dali-rp2
 ### Installation on Other Unix-like Systems
 * install python 3.8 or greater
 * install pip3
-* `pip install dali-rp2`
+
+Then install DaLI:
```
macanudo527, line 116:
```suggestion
Then install DaLI-RP2:
```

macanudo527 (approved):
Just a nitpick. I haven't had this issue myself running on Win 11, but it could help others.

README.md
jayr0d: There are other areas of README.md that say "Then install DaLI".  Would you like me to change them all?

eprbell: DaLI is the name of the software project, whereas dali-rp2 is the name of the Python package. I think in this context saying "Then install DaLI" is OK and shouldn't be too confusing. It's an English sentence directing the user to install the software, followed by the actual command to use in the CLI: I think the English sentence should use "DaLI" (which is less formal) and the command should (of course) use dali-rp2. Just my 2c.

macanudo527: Keep as is then. Thanks for the fixes!

PR #225, opened 2024-03-04
Implement forex python

### LINKED ISSUE #224: Forex Services Unavailable
Exchangerates.host has switched to a paid service. There aren't a lot of good free alternatives.

There are numerous places that we can download CSV data. For example [Forexsb](https://forexsb.com/historical-forex-data) offers detailed CSV files for free if you sign for a free account. I propose we modify the forex function to make use of CSVs as a backup.

What we could do is build a graph without prices and when missing prices are found, the user is alerted to download CSV files and put them in the cache folder. We can make use of more detailed information, but for now, only daily candles will be supported. That should be easy to work with since that is only around 5000 candles since 2008.

macanudo527:
## What?

Fixes #224 
Dependent on #223 
This allows users to make use of CSV files for Forex pricing instead of rely on exchangerate.host

## Testing

This has been tested with real world data and works.


macanudo527:
@eprbell this reads in missing forex data from CSV files. But, I did finally find[ this API](https://www.frankfurter.app/) that can basically do the same thing, but via a REST API. Should I keep both and use the CSV files as a fallback?

eprbell:
> @eprbell this reads in missing forex data from CSV files. But, I did finally find[ this API](https://www.frankfurter.app/) that can basically do the same thing, but via a REST API. Should I keep both and use the CSV files as a fallback?

Sounds good, but should it be a separate plugin, so that the user can decide which to use and with what priority?

macanudo527:
> Sounds good, but should it be a separate plugin, so that the user can decide which to use and with what priority?

I'm wondering about how to implement that. Maybe have two different CCXT pair converters one with the REST API, and one with the importing of CSV?

eprbell:
> > Sounds good, but should it be a separate plugin, so that the user can decide which to use and with what priority?
> 
> I'm wondering about how to implement that. Maybe have two different CCXT pair converters one with the REST API, and one with the importing of CSV?

Yes, that's what I was thinking: that should give the user the most flexibility.

macanudo527:
@eprbell btw, I'm going to try to cherry pick and splice the code from your PR. Hopefully, I can get it all put together here soon.

Should I implement this with inheritance, or by parameter that will load a particular class to handle forex processing? I'm thinking making ccxt pair converter abstract and the forex processing has to inherit from it. OR, make the REST API default, and then have another CSV class inherit and override the defaults.

eprbell:
> @eprbell btw, I'm going to try to cherry pick and splice the code from your PR. Hopefully, I can get it all put together here soon.

Sounds good, no worries.

> Should I implement this with inheritance, or by parameter that will load a particular class to handle forex processing? I'm thinking making ccxt pair converter abstract and the forex processing has to inherit from it. OR, make the REST API default, and then have another CSV class inherit and override the defaults.

I'm leaning toward making the ccxt-based converter abstract and have the two forex processing plugins inherit from it. In theory we could do something similar to the input plugins and have csv and rest subdirectories, but that would not be backward compatible, would require changes in users config files, and would make this change more complex: let's leave it as is. If needed, we'll tackle that issue as a separate change later.

macanudo527:
@eprbell This should fix most Forex Issues. It is lacking docs, but I thought I would let you get a peek at it.

eprbell:
> @eprbell This should fix most Forex Issues. It is lacking docs, but I thought I would let you get a peek at it.

Sounds good: I'll take a look (the diff in Github is large so it may take some time for me to go through it).

macanudo527:
> Sounds good: I'll take a look (the diff in Github is large so it may take some time for me to go through it).

I figured it would take a while. I tried to figure out a way to break down into simpler parts, but kind of just had to be done all at once.

src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -0,0 +1,1027 @@
...
+        result: List[HistoricalBar] = []
+        retry_count: int = 0
+        self.__transaction_count += 1
+        if timespan in _TIME_GRANULARITY_SET:
+            if exchange in _NONSTANDARD_GRANULARITY_EXCHANGE_SET:
+                retry_count = _TIME_GRANULARITY_DICT[exchange].index(timespan)
+            else:
+                retry_count = _TIME_GRANULARITY.index(timespan)
+        else:
+            raise RP2ValueError("Internal error: Invalid timespan passed to find_historical_bars.")
```
eprbell, line 491: typo: timespan -> time span?

src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -0,0 +1,1027 @@
...
+            if (timestamp - returned_bar.timestamp).total_seconds() > _TIME_GRANULARITY_STRING_TO_SECONDS[_ONE_DAY]:
+                raise RP2ValueError(
+                    "Internal error: The time difference between the requested and returned timestamps is "
+                    "greater than a day. The graph probably hasn't been optimized."
+                )
+            self._add_bar_to_cache(key=key, historical_bar=returned_bar)
+            return returned_bar
+        return None
+
+    def find_historical_bars(
```
eprbell, line 479: Opt/nit: this function is very large and could use some refactoring: consider breaking it into smaller bits in the future (you don't have to do it now, just keep it on your radar).

src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -0,0 +1,1027 @@
...
+            if alias_properties[0] in _EXCHANGE_DICT or alias_properties[0] == _UNIVERSAL:
+                exchange: str = alias_properties[0]
+            else:
+                raise RP2ValueError(f"Exchange - {alias_properties[0]} is not supported at this time. Check the spelling of the exchange.")
+
+            processed_aliases.setdefault(exchange, {}).update({Alias(alias_properties[1], alias_properties[2]): RP2Decimal(alias_properties[3])})
+
+        return processed_aliases
+
+    ### Abstract Fiat Methods (Must be overridden) ###
```
eprbell, line 989: I would just remove this comment: some of the methods below are not abstract, and the abstract one are clear from the raise clause in them.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Neal Chambers
+# Copyright 2024 Neal Chambers
```
eprbell, line 1: The file is not new, so the date should not change.

src/dali/plugin/pair_converter/ccxt.py
```diff
@@ -12,901 +12,254 @@
...
+
+                # This is a sanity check to make sure the format hasn't changed and that the request was successful.
+                if data[_AMOUNT] == 1.0:
+                    market: str = f"{from_asset}{to_asset}"
+                    rates: Dict[str, Any] = data[_RATES]
+                    previous_result: Optional[HistoricalBar] = None
+
+                    for day in range((end_of_year - beginning_of_year).days + 1):
+                        current_day: datetime = beginning_of_year + timedelta(days=day)
+                        print(f"current_day: {current_day}, timestamp: {key.timestamp}")
```
eprbell, line 129: Use the logger module instead of print.

src/dali/plugin/pair_converter/ccxt_fiat_from_csv.py
```diff
@@ -0,0 +1,120 @@
+# Copyright 2024 Neal Chambers
```
eprbell, line 1: Shouldn't this file live in the csv subdirectory?

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -62,7 +62,7 @@
 _GOOGLE_DRIVE_DOWNLOAD_URL: str = "https://drive.usercontent.google.com/download"
 
 # File ID for the unified CSV file ID. This will need to be replaced every quarter.
-_UNIFIED_CSV_FILE_ID: str = "16YKyFkYlvawCHv3W7WuTFzM8RYgMRWMt"
+_UNIFIED_CSV_FILE_ID: str = "11WtjXA9kvVYV9KDoebGV5U75dmcA3bJa"
```
eprbell, line 65: Can you add the full URL to the file into a comment, so that people can download it manually and inspect it, if they wish to do so?

tests/test_plugin_ccxt.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Neal Chambers
+# Copyright 2024 Neal Chambers
```
eprbell, line 1: No need to change the date.

eprbell (changes requested):
No worries, here's a first review.

src/dali/abstract_ccxt_pair_converter_plugin.py
macanudo527: Yeah 😄 There is a lot that needs to be refactored. I think I will spend more time on the core after this. :)

macanudo527:
@eprbell I tested this with real world data. It appears to be working as expected.

I had to implement 'pruning' because there were numerous hidden bugs that kept cropping up and the pruning process makes routing more brittle, but clearly shows us where the problems are. It also speeds up the snapshot / cloning process.

Basically, the unoptimized graph is pruned with the first set of optimizations before cloning starts. Edges are added and removed based on actual volume at the different snapshot times. I kept getting phantom routes that led to bad pricing and this resolves that. 

I'll try to clarify more of the changes when I have time.

eprbell:
> @eprbell I tested this with real world data. It appears to be working as expected.
> 
> I had to implement 'pruning' because there were numerous hidden bugs that kept cropping up and the pruning process makes routing more brittle, but clearly shows us where the problems are. It also speeds up the snapshot / cloning process.
> 
> Basically, the unoptimized graph is pruned with the first set of optimizations before cloning starts. Edges are added and removed based on actual volume at the different snapshot times. I kept getting phantom routes that led to bad pricing and this resolves that.
> 
> I'll try to clarify more of the changes when I have time.

Sounds good: I'll do another round of review.

src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -884,7 +901,7 @@ def _optimize_assets_for_exchange(
                     # market becomes available. Later, when the price is retrieved, the timestamps won't match and the user will be warned.
                     no_market_padding: HistoricalBar = HistoricalBar(
                         duration=bar_check[0].duration,
-                        timestamp=bar_check[0].timestamp - timedelta(weeks=1),
+                        timestamp=bar_check[0].timestamp - timedelta(weeks=4),  # Make this a parameter users can set?
```
eprbell, line 904: Let's use a constant instead of a hardcoded literal.

src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -986,7 +1006,9 @@ def _process_aliases(self, aliases: str) -> Dict[str, Dict[Alias, RP2Decimal]]:
 
         return processed_aliases
 
-    ### Abstract Fiat Methods (Must be overridden) ###
+    def _get_previous_monday(self, date: datetime) -> datetime:
+        days_behind = (date.weekday() + 1) % 7
```
eprbell, line 1010: The 7 should be a constant.

src/dali/mapped_graph.py
```diff
@@ -164,17 +180,43 @@ def clone_with_optimization(self, optimization: Dict[str, Dict[str, float]]) ->
...
                 else:
                     cloned_mapped_graph.add_vertex_if_missing(original_vertex.name)
 
-            # Add new neighbors
-            for optimized_asset, neighbor_weights in optimization.items():
-                for neighbor_name in neighbor_weights.keys():
-                    if original_vertex.name == optimized_asset:
-                        cloned_mapped_graph.add_neighbor(original_vertex.name, neighbor_name, neighbor_weights[neighbor_name], True)
+        # Add new neighbors
+        for optimized_asset, neighbor_weights in optimization.items():
```
eprbell, line 190: Could you explain why this for has been removed from the for at line 166? It's a bit unclear to me.

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -284,28 +285,48 @@ def _split_chunks_size_n(self, file_name: str, csv_file: str, chunk_size: int =
...
+                adjusted_chunk = [row for row in chunk if datetime.fromtimestamp(int(row[self.__TIMESTAMP_INDEX]), timezone.utc) >= next_monday]
+
+                i = 0
+                while i < len(adjusted_chunk):
+
+                    # When there is no volume for a day, Kraken doesn't create a row for that day
+                    # So we have to find 1-7 rows that are less than or equal to a week from the start of the week
+                    following_monday = next_monday + timedelta(days=7)
+                    week_of_chunks = [
+                        row for row in adjusted_chunk[i : i + 7] if datetime.fromtimestamp(int(row[self.__TIMESTAMP_INDEX]), timezone.utc) < following_monday
```
eprbell, line 304: 7 -> constant

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -365,23 +390,23 @@ def _retrieve_cached_bars(
                     with gopen(file_path, "rt") as file:
                         rows = reader(file)
                         for row in rows:
-                            if all_bars and int(row[self.__TIMESTAMP_INDEX]) >= duration_timestamp:
+                            if all_bars and int(float(row[self.__TIMESTAMP_INDEX])) >= duration_timestamp:
```
eprbell, line 393: Why the double cast (float + int)?

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -365,23 +390,23 @@ def _retrieve_cached_bars(
...
                                 result.append(
                                     HistoricalBar(
                                         duration=timedelta(minutes=int(_KRAKEN_TIME_GRANULARITY[retry_count])),
-                                        timestamp=datetime.fromtimestamp(int(row[self.__TIMESTAMP_INDEX]), timezone.utc),
+                                        timestamp=datetime.fromtimestamp(int(float(row[self.__TIMESTAMP_INDEX])), timezone.utc),
```
eprbell, line 397: Also, double cast. This is repeated several times in the following lines as well.

src/dali/plugin/pair_converter/csv/kraken.py
```diff
@@ -513,3 +538,9 @@ def _remove_unified_csv_file(self) -> None:
             self.__logger.info("%s has been safely deleted.", self.__UNIFIED_CSV_FILE)
         except FileNotFoundError:
             self.__logger.info("File %s not found.", self.__UNIFIED_CSV_FILE)
+
+    def _get_next_monday(self, date: datetime) -> datetime:
+        days_ahead = (7 - date.weekday()) % 7
```
eprbell, line 543: 7-> constant

eprbell (changes requested):
Here's a review of the latest changes: mostly minor stuff. After this is fixed, I'll do one more full review and then this can be merged, I think.

src/dali/mapped_graph.py
macanudo527: You mean why it is no longer in the for block starting at line 166?

The goal of this for loop is to add any optimizations that were not found in the graph. In the preceding for loop an **edge** optimization that is currently in the graph is popped off the dict in line 175:
https://github.com/eprbell/dali-rp2/blob/508dc43cbfc0bd99dfef85c522f3cecb6055df9d/src/dali/mapped_graph.py#L175

so, the only remaining **edges** in the optimization dict are ones that are not in the current graph and must be added. This loop goes through the leftover edges and applies them to nodes/points that are in the graph. 

Previously, it was indented because the graph had all the edges and so it would search for and replace the edges. Now the edges are deleted (not the nodes or points). so, the edges need to be added back in.

src/dali/plugin/pair_converter/csv/kraken.py
macanudo527: One of my files had a decimal point in the timestamp for some reason and I was getting an error with int. Is there a better way?

src/dali/plugin/pair_converter/csv/kraken.py
macanudo527: Yeah, the data had decimals in the timestamp. I can try to work on this in another PR and clean it up there. I honestly just wanted to get to an MVP with this.

src/dali/plugin/pair_converter/csv/kraken.py
eprbell: Does casting directly to int work?

macanudo527: nevermind, I found the kink in the line. When emulating the 1 week candle, the timestamp that is returned in line 312 is a float:
https://github.com/eprbell/dali-rp2/blob/508dc43cbfc0bd99dfef85c522f3cecb6055df9d/src/dali/plugin/pair_converter/csv/kraken.py#L308
I cast it as an int, then a str. I'm running real world data now to see if that fixes the issue.

macanudo527: The casting fixed it. I guess I got a little lazy there.

docs/configuration_file.md
```diff
@@ -528,6 +527,49 @@ Accuracy will improve once new CSV data is released, which is typically 2 weeks
...
+fiat_priority = <em>&lt;fiat_priority&gt;</em>
+untradeable_assets = <em>&lt;untradeable_assets&gt;</em>
+aliases = <em>&lt;untradeable_assets&gt;</em>
+</pre>
+
+Check the [CCXT plugin](#ccxt) section above for more information on the parameters.
+
+#### Note on Forex CSV file
+
+If you do not want to use Exchangerate.host or the Frankfurter API for Forex prices, you can use CSV files by saving them to the folder `.dali_cache/forex`. The files must be in the format of quote_asset_base_asset.csv (e.g. USD_JPY.csv). The format should be Time, high, low, open, close, volume. Like the below example:
```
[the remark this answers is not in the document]
eprbell: This seems unchanged.

docs/configuration_file.md
```diff
@@ -528,6 +527,49 @@ Accuracy will improve once new CSV data is released, which is typically 2 weeks
...
+
+Time               |Open  |High  |Low   |Close |Volume
+-------------------|------|------|------|------|------
+2008-02-22 00:00:00|210.45|215.45|211.56|213.45|8900
+
+Historical CSV Files are available from several places on the web for free. [Forexsb.com](https://forexsb.com/historical-forex-data) is a good source. Currently only 1 day candles are supported. One row per day a rate is needed is required. For example, if you want to use the [official yearly rates from the IRS](https://www.irs.gov/individuals/international-taxpayers/yearly-average-currency-exchange-rates), you will need to copy and paste the same rate for all days of the calendar year. CSV files can contain multiple years of rates.
```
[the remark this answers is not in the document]
eprbell: Also unchanged.

eprbell (changes requested):
I noticed a couple of conversation have been marked "resolved" but without fix. Once all the code is in: I'll do one last full review.

macanudo527:
> I noticed a couple of conversation have been marked "resolved" but without fix. Once all the code is in: I'll do one last full review.

@eprbell Sorry, I must have trampled it somewhere.

eprbell:
> > I noticed a couple of conversation have been marked "resolved" but without fix. Once all the code is in: I'll do one last full review.
> 
> @eprbell Sorry, I must have trampled it somewhere.

No worries!

eprbell (approved):
LGTM. I think this code reached a level of complexity that might benefit from some refactoring in terms of smaller functions (at some point in the future or as a background task): this would make it easier to read/maintain. Regardless, this is an impressive DaLI component: it went a long way since we started brainstorming about it a few years back. Great work!

macanudo527:
Next order of business is to add a new suite of tests to make it more robust. Then, I can refactor with a little less worry I'm breaking something, but yeah, feels good and the new pruning system speeds things up.

PR #228, opened 2024-03-25
Binance.US Implementation

tpaynter4:


src/dali/plugin/input/rest/binance_us.py
```diff
@@ -0,0 +1,1329 @@
...
+        )
+
+    @property
+    def _client(self) -> binanceus:
+        super_client: Exchange = super()._client
+        if not isinstance(super_client, binanceus):
+            raise RP2RuntimeError("Exchange is not instance of class binanceus.")
+        return super_client
+
+    def _get_algos(self) -> List[str]:
```
macanudo527, line 215: This function is associated with mining so it can be deleted too if Binance.us doesn't deal with mining.

src/dali/plugin/input/rest/binance_us.py
```diff
@@ -0,0 +1,1329 @@
...
+        api_key: str,
+        api_secret: str,
+        native_fiat: str,
+        username: Optional[str] = None,
+        thread_count: Optional[int] = __DEFAULT_THREAD_COUNT,
+    ) -> None:
+        self.__api_key = api_key
+        self.__api_secret = api_secret
+        # We will have a default start time of July 13th, 2017 since Binance Exchange officially launched on July 14th Beijing Time.
+        super().__init__(account_holder, datetime(2017, 7, 13, 0, 0, 0, 0), native_fiat, thread_count)
```
macanudo527, line 173: Can you track down the start date for Binance.us?

src/dali/plugin/input/rest/binance_us.py
```diff
@@ -0,0 +1,1329 @@
+# Copyright 2022 macanudo527
```
macanudo527, line 1: Be sure to take credit for this.

tests/test_plugin_binance_us.py
```diff
@@ -0,0 +1,893 @@
...
+        assert crypto_withdrawal_transaction.asset == "PAXG"
+        assert int(parser.parse(crypto_withdrawal_transaction.timestamp).timestamp()) * 1000 == 1599621997000
+        assert crypto_withdrawal_transaction.to_exchange == Keyword.UNKNOWN.value
+        assert crypto_withdrawal_transaction.from_exchange == "Binance.us"
+        assert crypto_withdrawal_transaction.crypto_received == Keyword.UNKNOWN.value
+        assert RP2Decimal(crypto_withdrawal_transaction.crypto_sent) == RP2Decimal("0.00999800")
+
+
+"""
+    def test_gains(self, mocker: Any) -> None:
```
macanudo527, line 451: if you want to disable a test, you can simply put disabled in front of the method name:
```suggestion
    def disabled_test_gains(self, mocker: Any) -> None:
```

macanudo527: Of course, if you never need gains, you can just delete the test entirely. You can always copy and paste the code from binance in the future if you think you need it.

macanudo527 (changes requested):
I think this is on the right course. I just remembered, we do have [exchange locked aliases that you could try to implement](https://github.com/eprbell/dali-rp2/blob/2096a7c62cb58dbcf7b1a9198da4aa998c80ebe7/src/dali/mapped_graph.py#L39-L50). 

You could also give that a try. Sorry I didn't think of it earlier. Since USD and USD4 are one-to-one, you can follow the example of the ETH2 -> ETH alias on Coinbase.

src/dali/plugin/input/rest/binance_us.py
tpaynter4: How exact do I need to be? I think it is 09/22/2019 to the nearest day.

tpaynter4:
I just realized that the plugin does not load my deposits or withdrawals. Are deposits and withdrawals processed in the _process_fiat_order method?

macanudo527:
> I just realized that the plugin does not load my deposits or withdrawals. Are deposits and withdrawals processed in the _process_fiat_order method?

They should be covered by the abstract plugin. Something might be run with the pagination details. You'll have to look into it. Is it missing fiat deposits and withdrawals only?

tpaynter4:
I believe it is only fiat.

macanudo527:
> I believe it is only fiat.

OK, well, my guess is that there is an implicit API function that covers fiat withdrawal/deposit you will have to look up. This looks like it right here:
https://docs.binance.us/#get-fiat-deposit-history

And the withdrawal Endpoint is there too.

tpaynter4:
The sapiGetFiatpaymentQueryWithdrawHistory and sapiGetFiatpaymentQueryDepositHistory end points are not available in the version of ccxt that dali-rp2 uses (3.0.79), but they are available in the current version (4.2.86). Is it possible to upgrade? Or would there be a way for me to use them with the current version?

macanudo527:
Yeah, that was on the todo list. Upgrades tend to break things. You can give it a try and I'll try with my exchanges as well, but these days I don't use as many advanced features.  

Be sure to double-check all your math. If something doesn't look right, there is probably a bug. CCXT is pretty finicky.

src/dali/plugin/input/rest/binance_us.py
macanudo527: An safe approximation is okay. Err on the side of being too early. For example, if you think it might be 9/22/2019 EST, you might as well make it 9/21/2019 just to be safe.

This just helps with optimization, so that we don't pull unnecessary information.

savgrace:
Is this ready to merge? I'm in need of a plugin for binance.us. If there's anything left to do I'd be happy help get this across the finish line with a few pointers :)

eprbell:
@savgrace, thanks for offering to help (sorry, this fell through the cracks and I just noticed your message now). @tpaynter4 and @macanudo527, is there any way @savgrace can help with this?

tpaynter4:
I've been pretty busy lately, but I am still having problems with deposits and withdrawals. I think there are pagination issues when using the CCXT endpoints for deposits and withdrawals, but I am not sure.

macanudo527:
> I've been pretty busy lately, but I am still having problems with deposits and withdrawals. I think there are pagination issues when using the CCXT endpoints for deposits and withdrawals, but I am not sure.

Have you looked at the API feed with Postman? In my experience, that has been a huge help for troubleshooting what the feed looks like.

PR #231, opened 2024-04-05
Make in, out, and intra CSV files optional

jayr0d:
Having in, out, and intra CSV files optional is helpful to new users that are using the manual CSV files for input.  They might start with one of the three and populate the rest later.  That is, they might make sure in transactions are processed successfully, then move on to out transactions.  When using an exchange as a qualifier, it is also possible there are only in transactions, as an example.  In this case, this change allows the user to use an ini file that does not have "out_csv_file = " and "intra_csv_file = " defined since there are only in transactions.


eprbell (approved):
LGTM.

PR #243, opened 2024-06-17
added Ireland plugin

advidb:


eprbell:
The static analysis errors in this PR will disappear once we merge the IE plugin in RP2 and I make a new RP2 release.

PR #249, opened 2024-07-05
Coinbase Pro intigration broken. Update configuration_file.md

7--:


macanudo527:
@eprbell Have you used the Coinbase Pro rest plugin? Does it still work? Coinbase Pro shutdown right? Shouldn't this plugin just be deleted?

eprbell:
> @eprbell Have you used the Coinbase Pro rest plugin? Does it still work? Coinbase Pro shutdown right? Shouldn't this plugin just be deleted?

I think CBPro REST endpoints have been [fully retired](https://docs.cdp.coinbase.com/advanced-trade/docs/faq#what-has-happened-to-coinbase-pro), so you're right that the plugin should just be deleted.

macanudo527:
Closing this because of the sunsetting of CoinbasePro.

PR #261, opened 2024-09-25
Disable Too Many Positional Arguments

macanudo527:
## What?
Pylint wants to punish us for too many positional arguments. I don't think we need this linter.

I also ran Black to add/delete some spaces.


macanudo527:
I pushed the wrong branch.

Issue #265, opened 2024-10-23
Implement Coinbase Advanced REST Input Plugin


macanudo527:
The CoinbasePro API has now been retired. The current input plugin needs to be deleted and a new Coinbase Advanced Plugin needs to be written. The [Python library](https://github.com/coinbase/coinbase-advanced-py) can be used or it can leverage [the CCXT plugin infastructure](https://github.com/eprbell/dali-rp2/blob/main/README.dev.md#data-loader-plugin-development).

- [ ] Remove CoinbasePro Plugin and accompanying documentation
- [ ] Add Coinbase Advanced Data Loader Plugin

G4te-Keep3r:
both Coinbase and pro get removed, right? Both are in table of contents in config file doc but pro has a dead link

G4te-Keep3r:
if the plan is to go all ccxt there could be a problem. As pip warns if you update ccxt
`dali-rp2 0.6.12 requires ccxt==3.0.79, but you have ccxt 4.4.63 which is incompatible`
With updated ccxt you can connect to coinbase fine via ccxt lib, but with ccxt 3.0.79 it throws same unauthorized error as dali-rp2 currently does with coinbase.

If you update ccxt to 4.4.63 dali-rp2 errors `cannot import name 'okex' from 'ccxt'`

I'll look into using cb python lib until ccxt versioning can be figured out.


Unless i'm having a user error with current coinbase dali-rp2 plugin, we need to remove all coinbase and coinbase pro and replace it with coinbase advanced only. At that point there is only 1 coinbase so does it still need the "advanced" descriptor?

macanudo527:
Basically, we need to test to see if the current version of CCXT works. The reason why we limited the version was because there was a bug introduced in the Binance part of it. I no longer use Binance, so I haven't gotten around to testing it. 

If you have the time, and you trade a few different exchanges, you could test to see upgrading to CCXT changes the data you get back by running dali-rp2 with your current data. Then, upgrade CCXT to the latest and run with your data again and see if you get the same results. That is all that really needs to be done. Then we can go ahead and bump up the CCXT version number. That would help us out a lot.

G4te-Keep3r:
seems okex renamed to okx, which will need to be updated in abstract_ccxt_pair_converter_plugin.py to match whenever the ccxt upgrade.

[...]

PR #266, opened 2024-10-29
Delete CoinbasePro Plugin

macanudo527:
## What?

Helps #265

The CoinbasePro Plugin no longer works since CoinbasePro is no longer available. This PR removes it so that users don't try to use it


macanudo527:
@eprbell I was going to submit this, but I think it can be revised with the Coinbase Advanced Python API. I'll have to look at it again.

macanudo527:
> @eprbell I was going to submit this, but I think it can be revised with the Coinbase Advanced Python API. I'll have to look at it again.

Sorry disregard this. This should probably be rewritten using the CCXT plugin so that it can all be standardized. We do need to rework the historical price plugin so that we have a spare pricing mechanism though.

eprbell (approved):

PR #269, opened 2024-11-19
Refactor of _optimize_assets_for_exchange

macanudo527:
## What?
This is essentially a refactoring of `_optimize_assets_for_exchange` 

I broke it into 6 stages and created individual functions for most of them to test them individually. This resulted in finding two bugs.

## Questions

I'm not sure if I should make these individual stage functions private or extract this logic into a separate module of some kind.


src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -853,126 +866,125 @@ def _get_request_delay(self, exchange: str) -> float:
...
-                        volume=bar_check[0].volume,
-                    )
-                    bar_check = [no_market_padding] + bar_check
-
-                    child_bars[child_name][neighbor.name] = bar_check
-                    timestamp_diff: float = (child_bars[child_name][neighbor.name][0].timestamp - start_date).total_seconds()
-
-                    # Find the start of the market if it is after the first transaction
-                    if timestamp_diff > _TIME_GRANULARITY_STRING_TO_SECONDS[_ONE_WEEK]:
-                        market_starts[child_name][neighbor.name] = child_bars[child_name][neighbor.name][0].timestamp
```
macanudo527, line 923: This was causing the `no_market_padding` bar to never be used, so I removed it.

macanudo527:
@eprbell Most of the code was simply moved into individual functions except the two pieces I commented on above.

eprbell:
> @eprbell Most of the code was simply moved into individual functions except the two pieces I commented on above.

Sounds good: I'm planning to review this weekend.

eprbell (approved):
Very nice: this is good refactoring and I love that we can unit test the new granular functions. We should strive to do this everywhere functions are bigger than 100 lines or so.

macanudo527:
> Very nice: this is good refactoring and I love that we can unit test the new granular functions. We should strive to do this everywhere functions are bigger than 100 lines or so.

Yeah, I'd like to do this for the core functions. It is a complicated beast and I found two bugs just in this one function.

eprbell (approved):

PR #270, opened 2024-12-04
Refactor find historical bars

macanudo527:
## What?

This is a refactor of `find_historical_bars`. I noticed while refactoring that pulling the bars from cache should be a guard clause at the top of the function. That caused a lot of revisions so I thought I would submit this early to make it easier to review.


macanudo527:
@eprbell Can you take a look at this when you get a chance?

eprbell:
> @eprbell Can you take a look at this when you get a chance?

Sure, I'll take a look in the weekend: could you fix the CI failure before that?

tests/test_abstract_ccxt_pair_converter.py
```diff
@@ -74,10 +81,11 @@ def unoptimized_graph(self, vertex_list: Dict[str, Vertex[str]]) -> MappedGraph[
 
     @pytest.fixture
     def historical_bars(self) -> Dict[str, HistoricalBar]:
+        now_time = datetime.now(timezone.utc)
         return {
             MARKET_START: HistoricalBar(
                 duration=timedelta(weeks=1),
-                timestamp=datetime(2023, 1, 1),
+                timestamp=now_time,
```
eprbell, line 88: Since this is a test wouldn't it be better to have fixed data (like 2023/1/1) rather than datetime.now()?

eprbell (changes requested):
Looks good to me, after CI is fixed.

src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -513,16 +514,23 @@ def find_historical_bar(self, from_asset: str, to_asset: str, timestamp: datetim
     def find_historical_bars(
         self, from_asset: str, to_asset: str, timestamp: datetime, exchange: str, all_bars: bool = False, timespan: str = _MINUTE
     ) -> Optional[List[HistoricalBar]]:
+
+        # Guard clause to pull from cache
+        if all_bars:
+            cached_bundle = self._get_bundle_from_cache(AssetPairAndTimestamp(timestamp, from_asset, to_asset, exchange))
+            if cached_bundle:
+                # If the last bar in the bundle is within the last week, return the bundle
+                if (datetime.now(timezone.utc) - cached_bundle[-1].timestamp).total_seconds() <= _TIME_GRANULARITY_STRING_TO_SECONDS[_ONE_WEEK]:
```
macanudo527, line 523: @eprbell I don't think I can use a set time for the example bars, because this guard clause checks for stale data by checking if the stamp is within the last week.

tests/test_abstract_ccxt_pair_converter.py
macanudo527: I don't think I can since the guard clause checks for stale data. It needs to be a timestamp within the last week. See other comment.

macanudo527:
@eprbell I had to downgrade `urllib3` due to an incompatibility with vcrpy 4.4 and python 3.10. I think we need to drop support for Python 3.8 and 3.9. I can do that in another PR.

macanudo527:
@eprbell  I also removed the pandas stubs we wrote since there is a package that covers those now.

eprbell:
> @eprbell I had to downgrade `urllib3` due to an incompatibility with vcrpy 4.4 and python 3.10. I think we need to drop support for Python 3.8 and 3.9. I can do that in another PR.
> 
> If we drop support for 3.8/3.9 we can use the latest versions of vcrpy and urllib3.

Rather than downgrading libraries, I think we should change the supported Python to 3.10...3.13: no need to keep supporting old versions if they force us to use old dependencies.

setup.cfg
```diff
@@ -46,6 +47,7 @@ install_requires =
     pytz>=2021.3
     requests>=2.26.0
     rp2>=1.7.1
+    urllib3<2 # 2+ is incompatible with Python ~3.9
```
eprbell, line 50: Let's instead support Python 3.10...3.13.

eprbell (changes requested):
LGTM: let's just update supported Python versions.

setup.cfg
macanudo527: OK, should be good to go now. Thanks.

eprbell (approved):
LGTM

PR #283, opened 2025-01-23
Remove Unnecessary Alt Market

macanudo527:
## What?

When an alternative market exists before a perfect match market (e.g. TAO->USDT exists before TAO->USD) we need to force routing through the system to use the alternative market. Once the main market becomes available we need to stop using the alternative market. Also, the padding that is automatically applied to markets before it starts trading on an exchange needs to be deleted when an alternative market exists that can provide unpadded prices.

This code looks for assets that are already optimized that are in alternative markets. If it finds that the main market comes available it will delete the padding automatically applied and use the alternative market. Once the market actually starts it will delete subsequent alternative markets so that there is no overlap.

### DIFF
--- a/src/dali/abstract_ccxt_pair_converter_plugin.py
+++ b/src/dali/abstract_ccxt_pair_converter_plugin.py
@@ -246,6 +246,9 @@ DEFAULT_FIAT_LIST: List[str] = ["AUD", "CAD", "CHF", "EUR", "GBP", "JPY", "NZD",
 # How much padding in weeks should we add to the graph to catch airdropped or new assets that don't yet have a market
 MARKET_PADDING_IN_WEEKS: int = 4
 
+DELETABLE_VOLUME: float = -1.0
+PADDING_VOLUME: float = -2.0
+
 DAYS_IN_WEEK: int = 7
 MANY_YEARS_IN_THE_FUTURE: relativedelta = relativedelta(years=100)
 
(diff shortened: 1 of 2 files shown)


src/dali/abstract_ccxt_pair_converter_plugin.py
```diff
@@ -1017,8 +1020,63 @@ def _refine_and_finalize_optimizations(self, optimizations: Dict[datetime, Dict[
...
+                        # If all the neighbors are padding, delete the asset if an alternative market exists
+                        if all(volume == PADDING_VOLUME for _, volume in neighbors.items()):
+                            self._logger.debug("Deleting Padding %s with %s", asset, neighbors.keys())
+                            timestamp_assets_to_delete.setdefault(timestamp, []).append(asset)
+                        else:
+                            deletable_market[asset] = alternative_market
+                    elif asset in deletable_market and deletable_market[asset] == current_neighbors[0]:
+                        self._logger.debug("Deleting alternative market for %s at %s", asset, timestamp)
+                        # Note this deletes the asset from the snapshot and all its neighbors not just the alternative market
+                        # This will work in most cases, but if the asset has multiple alternative markets, it will delete all of them
```
eprbell, line 1072: Nit: add a period after "delete all of them"

eprbell (approved):
Just one nit in a comment. LGTM

PR #284, opened 2025-01-25
Update bitbank supplemental CSV

macanudo527:
## What??

Bitbank.cc updated their CSVs to include all assets, so you only have to export one withdrawal and one deposit for crypto assets and a withdrawal/deposit files for yen. This update allows you to use the new format.


eprbell (approved):
LGTM.

PR #293, opened 2025-11-11
feat: Create coinbase-advanced rest plugin

randumari:
Adding a new coinbase_advanced rest plugin.  The old plugins are no longer functional, as Coinbase Pro no longer exists and the old rest api requires different authentication.

This also includes a change to the "coinbase_advanced" spot price retrieval plugin, as it was failing due to Coinbase's introduction of USDC without providing a USDC/USD product-id/lookup.

A Dockerfile is included in case security-conscious users want to run in a sandbox and also because the new rest plugin requires the Coinbase Rest Client, which this Dockerfile auto-installs.

### DIFF
--- a/.github/workflows/documentation_check.yml
+++ b/.github/workflows/documentation_check.yml
@@ -1,11 +1,22 @@
 name: Documentation Check
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
 
 jobs:
   markdown-link-check:
     name: Check markdown links
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@main
-    - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v2
+        with:
+          # Check all markdown files recursively
+          args: --config lychee.toml --verbose --no-progress './**/*.md'
+          # Fail the job if broken links are found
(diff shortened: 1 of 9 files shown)


randumari:
What else is needed to get this merged in?

I see Documentation Check errors with broken links, but those links don't appear to be related to changes in this PR.

I believe I've fixed all errors reported by pylint.  Some warnings and convention level messages remain and I could resolve a few of those, but a few I disagree with, too.  Does this project treat pylint warnings as a cause for static analysis failure, or are there errors that I didn't see in the log?

anlach:
Were there changes to abstract_input_plugin.py that are not included? This doesn't work for me.

```
ERROR: Fatal exception occurred:
Traceback (most recent call last):
  File "/home/anlach/Code/dali-rp2/src/dali/dali_main.py", line 141, in _dali_main_internal
    input_plugin: AbstractInputPlugin = plugin_module.InputPlugin(**plugin_configuration)
  File "/home/anlach/Code/dali-rp2/src/dali/plugin/input/rest/coinbase_advanced.py", line 179, in __init__
    self.__api_url: str = InputPlugin.__API_URL
AttributeError: type object 'InputPlugin' has no attribute '_InputPlugin__API_URL'
```

randumari:
@anlach Thank you for reporting this.  I'm not sure how this was working before, but I reproduced that issue.  It seems to be some stray code that I copied from the original Coinbase plugin and isn't needed for the coinbase-advanced plugin, so I removed it (along with another typo).  The changes are now in this PR.

